PR #18, opened 2025-10-27
Add build_test file for RCCL (#267)

sudharssun:
Summary:
As in title.

Also delete the missed setup_rccl.sh copy in github (previous diff)

Differential Revision: D84125652

Pulled By: sudharssun

### DIFF
--- a/.github/workflows/build_test_rccl.yaml
+++ b/.github/workflows/build_test_rccl.yaml
@@ -1,0 +1,72 @@
+name: Build RCCL
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


.github/workflows/build_test_rccl.yaml
```diff
@@ -0,0 +1,75 @@
...
+      contents: read
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: RCCL
+            permissions:
```
atalman: Please remove permissions from here

Issue #26, opened 2025-10-30
Question: Is it possible to get rid of `conda` when compiling?


Gin-Sin:
I noticed that `CMakeLists.txt` needs to find `CONDA_PREFIX` which is automatically exported if `conda` is installed. However, it is confusing why it is necessary to involve `conda` in this package. I understand it might be consistent with the development habits in your team, but I would like to make it portable to other environments, so `conda` is deemed too heavy under this situation. Is it possible to get rid of `conda`? What is the reason to involve `conda`? Thanks!

Gin-Sin:
> If you specify `USE_SYSTEM_LIBS=1` & `CONDA_PREFIX=/usr`, then torchcomms will use a custom version of nccl `libnccl.so` under `/usr/${LIB_SUFFIX}` (See [CMakeLists.txt](https://github.com/meta-pytorch/torchcomms/blob/main/comms/torchcomms/nccl/CMakeLists.txt#L8-L10)). As long as you have dependency glog==0.4.0 & fmt at system level. You can directly build the torchcomms from scratch with custom nccl backend as follows:
> 
> ```
> export USE_GLOO=0
> export USE_NCCLX=0
> export USE_NCCL=1
> export USE_SYSTEM_LIBS=1
> export CONDA_PREFIX=/usr
> pip install -v .
> ```

Is there any version constraint on `libfmt`? I successfully build `torchcomms` with just `NCCL`, preinstalling `libfmt==8.1.1`. However, when I tried to `import torchcomms`, the following errors popped out:

```
Traceback (most recent call last):
  File "/code/torchcomms/nccl_all_reduce.py", line 4, in <module>
    from torchcomms import new_comm, ReduceOp
  File "/usr/local/lib/python3.10/dist-packages/torchcomms/__init__.py", line 20, in <module>
    _load_libtorchcomms()
  File "/usr/local/lib/python3.10/dist-packages/torchcomms/__init__.py", line 17, in _load_libtorchcomms
    ctypes.CDLL(libtorchcomms_path, mode=ctypes.RTLD_LOCAL)
  File "/usr/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
[...]

Issue #21, opened 2025-10-28
libglog and libfmt linkage fails due to looking for static, not shared, libs


jeffhammond:
I followed the instructions exactly and the torchcomms build fails here in multiple environments.  Why is the build system trying to use the static library when the shared library has been installed by Conda?  What option do I need to set to change this?

```
  [ 27%] Linking CXX shared library comms/torchcomms/libtorchcomms.so
  /opt/conda/envs/torchcomms/bin/cmake -E cmake_link_script CMakeFiles/torchcomms.dir/link.txt --verbose=1
  /usr/bin/ld: cannot find -l:libglog.a: No such file or directory
  /usr/bin/ld: cannot find -l:libfmt.a: No such file or directory
  collect2: error: ld returned 1 exit status
[...]

PR #43, opened 2025-11-04
support conda free builds

d4l3k:
This supports conda free builds as well as just doing `pip install` with no `./build_ncclx.sh`.

Key changes:

* build openssl static from source instead of depending on conda
* if no CONDA_PREFIX is detected we use `build/conda` as a dummy conda prefix
* if ncclx isn't built we automatically call `./build_ncclx.sh`

Related issues: #21, #26 

Test plan:

```
uv venv --python 3.14
uv pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu128
uv pip install --no-build-isolation -v ../torchcomms

source .venv/bin/activate.fish
MASTER_ADDR=localhost MASTER_PORT=0 RANK=0 WORLD_SIZE=1 python -c "import torch; import torch.distributed as dist; import torchcomms; comm = torchcomms
.new_comm('ncclx', torch.device('cuda'), store=None, name='1234'); t = torch.zeros(10, device=comm.get_device()); comm.all_reduce(t, torchcomms.ReduceOp.SUM, async_op=False); comm.finalize()"
```


CMakeLists.txt
```diff
@@ -29,11 +29,15 @@ set(ROOT ${CMAKE_CURRENT_SOURCE_DIR})
...
+message(STATUS "Python3_ROOT_DIR : ${Python3_ROOT_DIR}")
+
 # Find CONDA_PREFIX
-if(NOT DEFINED ENV{CONDA_PREFIX})
-    message(FATAL_ERROR "CONDA_PREFIX environment variable not set")
+if(DEFINED ENV{CONDA_PREFIX})
+  set(CONDA_PREFIX $ENV{CONDA_PREFIX})
+else()
+  set(CONDA_PREFIX "${ROOT}/build/conda")
+  set(ENV{CONDA_PREFIX} "${CONDA_PREFIX}")
```
tanquer, line 41: If `CONDA_PREFIX` is not defined, should it be system level prefix `/usr`? I think we should change the name from `CONDA_PREFIX` to `PREFIX` and update the following `CONDA_LIB` to `LIB` and `CONDA_INCLUDE` to `INCLUDE`.

d4l3k: Yeah, you're right, definitely should clean this up. 

We don't want to use the system prefix as we install into it when building depa from source

tanquer: I see, then the logic might should be
```
if(DEFINED ENV{CONDA_PREFIX})
  set(CONDA_PREFIX $ENV{CONDA_PREFIX})
elif(NOT DEFINED ENV{USE_SYSTEM_LIBS})
  set(CONDA_PREFIX "${ROOT}/build/conda")
  set(ENV{CONDA_PREFIX} "${CONDA_PREFIX}")
else()
  ...
endif()
```

build_ncclx.sh
```diff
@@ -120,6 +141,7 @@ function build_third_party {
     build_fb_oss_library "https://github.com/fmtlib/fmt.git" "11.2.0" fmt "-DFMT_INSTALL=ON -DFMT_TEST=OFF -DFMT_DOC=OFF -DBUILD_SHARED_LIBS=ON"
     build_fb_oss_library "https://github.com/madler/zlib.git" "v1.2.13" zlib "-DZLIB_BUILD_TESTING=OFF"
     build_boost
+    build_openssl
```
tanquer, line 144: Can you also add `conda-forge::libopenssl-static` at line 173 for user specifying the `USE_SYSTEM_LIBS`?

CMakeLists.txt
d4l3k: Removing all CONDA_* prefixes is going to be a lot of changes. I'm going to do that as a follow up PR to keep things clean

build_ncclx.sh
d4l3k: done

CMakeLists.txt
tanquer: Using the system prefix is for user who install the deps by themselves. Their deps (ex. libnccl.so) are under `/usr/lib`. And if we define `CONDA_PREFIX = "${ROOT}/build/conda"`, then the following `CONDA_LIB` & `CONDA_INCLUDE` might be an empty/non-exist directory, then, we might not be able to link the correct dynamic library (ex. See [`comms/torchcomms/nccl/CMakeLists.txt:line 10`](https://github.com/meta-pytorch/torchcomms/blob/main/comms/torchcomms/nccl/CMakeLists.txt?fbclid=IwY2xjawN3zQJleHRuA2FlbQIxMQBicmlkETF2TWs0NENWRzJzTGV1SjFWc3J0YwZhcHBfaWQBMAABHkwSOaMoQ730_jdN33KciUJ10wfcR9hmLNRxBvfYh-GWeaNnrfTWMpAsS90__aem_yFPcF08XTeG11GaLwNLc1Q#L10)). Maybe we should change the `comms/torchcomms/nccl/CMakeLists.txt` to use `-lnccl` instead of `${NCCL_SHARED_LIB}`

Issue #51, opened 2025-11-10
RFC: Add XCCL Backend Support for Intel GPU in TorchComms


zhangxiaoli73:
## Summary

This RFC proposes adding XCCL (Intel GPU Collective Communications Library) backend support to TorchComms, enabling efficient distributed communication on Intel GPUs. This backend leverages Intel's oneCCL library and integrates with PyTorch's XPU device support via SYCL runtime.

---

## Motivation

Currently, TorchComms supports NCCL (NVIDIA), RCCL (AMD), and Gloo (CPU) backends, but lacks native support for Intel GPUs. Since PyTorch provides distributed support for Intel GPUs through the `xccl` backend, we plan to integrate XCCL into TorchComms as well.

### Architecture

The XCCL backend follows the same architectural pattern as existing backends (NCCL/RCCL):

```
Python API (torchcomms)
    ↓
TorchCommBackend Interface
    ↓
TorchCommXCCL Implementation
    ↓
    ├── XcclApi (oneCCL wrapper)
    ├── XpuApi (SYCL/XPU abstraction)
    └── TorchCommXCCL (Collectives implementation)
    
```

#### Core Files

| **File**                          | **Purpose**                         |
| --------------------------------- | ----------------------------------- |
| `TorchCommXCCL.cpp/.hpp`          | Collectives implementation    |
| `TorchWorkXCCL.cpp/.hpp`          | Asynchronous work tracking          |
| `TorchWorkXCCLQueue.cpp`          | Work queue management               |
| `TorchCommXCCLBootstrap.cpp/.hpp` | Communicator initialization         |
| `TorchCommXCCLUtils.cpp/.hpp`     | Utility functions (type conversion) |
| `TorchCommXCCLPy.cpp`             | Python bindings                     |

#### API Abstraction Layers

| **File**                 | **Purpose**                  |
| ------------------------ | ---------------------------- |
| `XcclApi.cpp/.hpp`       | oneCCL API wrapper           |
| `device/XpuApi.cpp/.hpp` | SYCL/XPU runtime abstraction |

#### Build System

| **File**              | **Purpose**                          |
| --------------------- | ------------------------------------ |
| `xccl/CMakeLists.txt` | CMake configuration for XCCL backend |
| `setup.py`            | Python package build (USE_XCCL flag) |

### API Surface

The XCCL backend implements all standard TorchComm operations:

#### Point-to-Point Operations

- `send()` / `recv()`
[...]

PR #52, opened 2025-11-10
Add XCCL Backend Support for Intel GPU in TorchComms

zhangxiaoli73:
### Motivation:

As design illustrated in RFC https://github.com/meta-pytorch/torchcomms/issues/51, we would like to add XCCL (Intel GPU Collective Communications Library) backend support to TorchComms. 

In this PR, we will enable `allreduce` in the XCCL backend as an entry point. Full collectives support will come later.

#### Core Files

| **File**                          | **Purpose**                         |
| --------------------------------- | ----------------------------------- |
| `TorchCommXCCL.cpp/.hpp`          | Main collectives implementation    |
| `TorchWorkXCCL.cpp/.hpp`          | Asynchronous work tracking          |
| `TorchWorkXCCLQueue.cpp`          | Work queue management               |
| `TorchCommXCCLBootstrap.cpp/.hpp` | Communicator initialization         |
| `TorchCommXCCLUtils.cpp/.hpp`     | Utility functions (type conversion) |
| `TorchCommXCCLPy.cpp`             | Python bindings                     |

#### API Abstraction Layers

| **File**                 | **Purpose**                  |
| ------------------------ | ---------------------------- |
| `XcclApi.cpp/.hpp`       | oneCCL API wrapper           |
| `device/XpuApi.cpp/.hpp` | SYCL/XPU runtime abstraction |

#### Build System

| **File**              | **Purpose**                          |
| --------------------- | ------------------------------------ |
| `xccl/CMakeLists.txt` | CMake configuration for XCCL backend |
| `setup.py`            | Python package build (USE_XCCL flag) |
 

### Examples:
```python
import torch
import torchcomms

# Initialize communicator
device = torch.device('xpu')
comm = new_comm("xccl", device, name="main_comm")
rank = torchcomms.get_rank()
current_device = torch.device(f"xpu:{rank}")

# Perform collective operation
tensor = torch.randn(1024, device=current_device)
work = comm.all_reduce(tensor, torchcomms.ReduceOp.SUM, async_op=True)
work.wait()

# Cleanup
comm.finalize()
```

### DIFF
--- a/comms/torchcomms/device/XpuApi.hpp
+++ b/comms/torchcomms/device/XpuApi.hpp
@@ -1,10 +1,10 @@
 #pragma once
 
-#include <sycl/sycl.hpp>
 #include <ATen/xpu/XPUContext.h>
 #include <ATen/xpu/XPUEvent.h>
-#include <c10/xpu/XPUStream.h>
 #include <c10/xpu/XPUFunctions.h>
+#include <c10/xpu/XPUStream.h>
+#include <sycl/sycl.hpp>
 
 namespace torch {
 namespace comms {
@@ -13,22 +13,22 @@ using xpuStream_t = ::c10::xpu::XPUStream;
 using xpuEvent_t = ::at::xpu::XPUEvent;
 
 struct xpuDeviceProp {
-    char name[256];
-    size_t totalGlobalMem;
-    int multiProcessorCount;
-    int maxThreadsPerBlock;
-    int maxThreadsDim[3];
+  char name[256];
+  size_t totalGlobalMem;
+  int multiProcessorCount;
+  int maxThreadsPerBlock;
+  int maxThreadsDim[3];
 };
 
 // Graph-related types (placeholder - unsupported in XPU)
 using xpuGraph_t = void*;
 using xpuGraphNode_t = void*;
 using xpuUserObject_t = void*;
-using xpuHostFn_t = void(*)(void*);
+using xpuHostFn_t = void (*)(void*);
 
 // Stream capture status (not supported in XPU)
 enum xpuStreamCaptureStatus {
-    xpuStreamCaptureStatusNone = 0,
+  xpuStreamCaptureStatusNone = 0,
 };
 
 // Error code type
@@ -40,15 +40,15 @@ constexpr xpu_result_t XPU_ERROR_INVALID_HANDLE = 3;
 constexpr xpu_result_t XPU_ERROR_OUT_OF_MEMORY = 4;
 constexpr xpu_result_t XPU_ERROR_UNSUPPORTED = 5;
 
-#define XPU_CHECK(xpu_api, call, err_str)                                 \
-  do {                                                                    \
-    xpu_result_t status = call;                                           \
-    if (status != XPU_SUCCESS) {                                          \
-      std::stringstream ss;                                               \
-      ss << err_str << ": " << xpu_api->getErrorString(status) << " at "  \
-         << __FILE__ << ":" << __LINE__;                                  \
-      throw std::runtime_error(ss.str());                                 \
-    }                                                                     \
+#define XPU_CHECK(xpu_api, call, err_str)                                \
+  do {                                                                   \
+    xpu_result_t status = call;                                          \
+    if (status != XPU_SUCCESS) {                                         \
+      std::stringstream ss;                                              \
+      ss << err_str << ": " << xpu_api->getErrorString(status) << " at " \
+         << __FILE__ << ":" << __LINE__;                                 \
+      throw std::runtime_error(ss.str());                                \
+    }                                                                    \
   } while (0)
 
 /**
@@ -101,7 +101,9 @@ class XpuApi {
       xpuEvent_t& event,
       unsigned int flags) = 0;
   virtual xpu_result_t eventDestroy(const xpuEvent_t& event) = 0;
-  virtual xpu_result_t eventRecord(xpuEvent_t& event, const xpuStream_t& stream) = 0;
+  virtual xpu_result_t eventRecord(
+      xpuEvent_t& event,
+      const xpuStream_t& stream) = 0;
   virtual xpu_result_t eventQuery(const xpuEvent_t& event) = 0;
 
   // Graph operations (unsupported, kept for API compatibility)
@@ -169,9 +171,11 @@ class DefaultXpuApi : public XpuApi {
 
   // Event management
   xpu_result_t eventCreate(xpuEvent_t& event) override;
-  xpu_result_t eventCreateWithFlags(xpuEvent_t& event, unsigned int flags) override;
+  xpu_result_t eventCreateWithFlags(xpuEvent_t& event, unsigned int flags)
+      override;
   xpu_result_t eventDestroy(const xpuEvent_t& event) override;
-  xpu_result_t eventRecord(xpuEvent_t& event, const xpuStream_t& stream) override;
+  xpu_result_t eventRecord(xpuEvent_t& event, const xpuStream_t& stream)
+      override;
   xpu_result_t eventQuery(const xpuEvent_t& event) override;
 
   // Graph operations (unsupported)
--- a/comms/torchcomms/xccl/TorchCommXCCL.cpp
+++ b/comms/torchcomms/xccl/TorchCommXCCL.cpp
@@ -1,17 +1,19 @@
 #include "comms/torchcomms/xccl/TorchCommXCCL.hpp"
 
-#include "comms/torchcomms/TorchCommFactory.hpp"
-#include "comms/torchcomms/TorchCommLogging.hpp"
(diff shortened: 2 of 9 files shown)


comms/torchcomms/xccl/TorchCommXCCLUtils.cpp
```diff
@@ -0,0 +1,314 @@
...
+    return onecclMin;
+  case ReduceOp::RedOpType::MAX:
+    return onecclMax;
+  case ReduceOp::RedOpType::BAND:
+    return onecclSum; // XCCL doesn't have bitwise AND, using SUM as fallback
+  case ReduceOp::RedOpType::BOR:
+    return onecclSum; // XCCL doesn't have bitwise OR, using SUM as fallback
+  case ReduceOp::RedOpType::BXOR:
+    return onecclSum; // XCCL doesn't have bitwise XOR, using SUM as fallback
+  case ReduceOp::RedOpType::PREMUL_SUM:
```
d4l3k: I'm concerned that this will cause silent corruption unless there's another piece to this that I'm missing. Can we throw an error for these?

comms/torchcomms/device/XpuApi.hpp
```diff
@@ -0,0 +1,202 @@
...
+      xpuUserObject_t object,
+      unsigned int count,
+      unsigned int flags) = 0;
+  virtual xpu_result_t streamGetCaptureInfo_v2(
+      const xpuStream_t& stream,
+      xpuStreamCaptureStatus* captureStatus_out,
+      unsigned long long* id_out,
+      xpuGraph_t* graph_out,
+      const xpuGraphNode_t** dependencies_out,
+      size_t* numDependencies_out) = 0;
```
d4l3k, line 127: Are you trying to keep this as close as possible to the NCCL impl? given we don't use this might as well just delete them. There's no virtual interface these are implementing afaik

comms/torchcomms/device/XpuApi.cpp
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+xpu_result_t DefaultXpuApi::streamCreateWithPriority(
+    xpuStream_t& stream,
+    unsigned int flags,
+    int priority) {
+    try {
+        // Map priority: priority < 0 = high, priority >= 0 = normal
+        bool isHighPriority = (priority < 0);
```
d4l3k: converting a bool to an int and back to a bool feels error prone

comms/torchcomms/xccl/TorchWorkXCCL.cpp
```diff
@@ -0,0 +1,130 @@
...
+    auto elapsed_milliseconds =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            current_time - start_completed_time_.value());
+
+    // Check if the operation has timed out
+    if (elapsed_milliseconds > timeout_ms_) {
+      // Operation has timed out
+      state_ = WorkStatus::TIMEDOUT;
+    }
+  } else if (end_status != XPU_ERROR_UNSUPPORTED) {
```
d4l3k: why does UNSUPPORTED not set an error state?

comms/torchcomms/xccl/XcclApi.cpp
d4l3k, line 1: General comment on this Api file: we use the API file for NCCL/NCCLX since we want to be able to directly test the logic in the torchcomm backend impl via mocks.

Are you planning on adding similar C++ in depth testing? I would recommend it, but if not, there's no real need for this abstraction layer

comms/torchcomms/xccl/TorchWorkXCCL.hpp
```diff
@@ -0,0 +1,95 @@
...
+  std::chrono::milliseconds timeout_ms_;
+
+  // state machine variables. TODO: convert to state machine later
+  std::atomic<WorkStatus> state_;
+
+  std::optional<std::chrono::steady_clock::time_point> start_completed_time_;
+  std::shared_ptr<TorchCommTracing> tracing_;
+};
+
+class TorchWorkXCCLQueue {
```
d4l3k, line 81: This seems really easy to templatize and share across all impls

comms/torchcomms/device/XpuApi.hpp
zhangxiaoli73: We do have a plan to enable XPU graphs in the XCCL backend, similar to CUDA graphs in the NCCL backend. Therefore, we keep these APIs in place and will add the implementations later.

comms/torchcomms/device/XpuApi.cpp
zhangxiaoli73: Thanks, good catch！ We used the API below, which required converting values from int to bool.
```
XPUStream getStreamFromPool(const bool isHighPriority, DeviceIndex device)
```
To avoid the conversion, we can use another API to pass priority directly.
```
XPUStream getStreamFromPool(const int priority, DeviceIndex device)
```

comms/torchcomms/xccl/XcclApi.cpp
zhangxiaoli73: Yes, we do have plan to enable C++ UT for XCCL backend to ensure better code quality.

comms/torchcomms/xccl/TorchWorkXCCL.hpp
zhangxiaoli73: Yes, it could be generalized to reduce code duplication. Would it be OK if I separate that generalization into another PR? As you know, this abstraction may also impact other backends like NCCL.

comms/torchcomms/xccl/TorchCommXCCLUtils.cpp
zhangxiaoli73: Make sense, it's very risky and let me throw exception here.

 Actually, what we did is very closely to NCCL backend (using SUM as fallback) https://github.com/meta-pytorch/torchcomms/blob/main/comms/torchcomms/nccl/TorchCommNCCLUtils.cpp#L150

comms/torchcomms/xccl/TorchWorkXCCL.cpp
zhangxiaoli73: You are right. We do need to throw error state here. Change done.

comms/torchcomms/xccl/TorchCommXCCL.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+// Collective Operations
+c10::intrusive_ptr<TorchWork>
+TorchCommXCCL::broadcast(at::Tensor &tensor, int root, bool async_op,
+                         const BroadcastOptions &options) {
+  throw std::runtime_error("XCCL broadcast is not supported now and will be added later");
+}
+
+c10::intrusive_ptr<TorchWork>
+TorchCommXCCL::all_reduce(at::Tensor &tensor, const ReduceOp &op, bool async_op,
```
d4l3k: Given you have implemented this, do you mind including in the test plan the test results for `AllReduceTest.py`?

zhangxiaoli73: Sure. I run this UT with `torchrun --nproc_per_node=4 --rdzv_endpoint=localhost:29501 AllReduceTest.py`. The full log is quite long. Pieces of test log like below, you can see 7 tests were run and 2 skipped(skipped ncclx-only tests), 5 passed.
```
TorchComm created with params: backend xccl, device xpu:0, options <torchcomms.CommOptions object at 0x7fc3a52de570>
Testing async all_reduce with input deleted after enqueue with count=0, dtype=Float, and op=Sum
Testing async all_reduce with input deleted after enqueue with count=0, dtype=Float, and op=Max
Testing async all_reduce with input deleted after enqueue with count=0, dtype=Int, and op=Sum
Testing async all_reduce with input deleted after enqueue with count=0, dtype=Int, and op=Max
Testing async all_reduce with input deleted after enqueue with count=0, dtype=SignedChar, and op=Sum
Testing async all_reduce with input deleted after enqueue with count=0, dtype=SignedChar, and op=Max
Testing async all_reduce with input deleted after enqueue with count=4, dtype=Float, and op=Sum
Testing async all_reduce with input deleted after enqueue with count=4, dtype=Float, and op=Max
Testing async all_reduce with input deleted after enqueue with count=4, dtype=Int, and op=Sum
Testing async all_reduce with input deleted after enqueue with count=4, dtype=Int, and op=Max

....

Testing sync all_reduce without work object with count=1048576, dtype=BFloat16, and op=PremulSum
Testing sync all_reduce without work object with count=1048576, dtype=BFloat16, and op=PremulSum
Testing sync all_reduce without work object with count=1048576, dtype=BFloat16, and op=PremulSum
...
----------------------------------------------------------------------
Ran 7 tests in 162.128s

OK (skipped=2)

----------------------------------------------------------------------
```

zhangxiaoli73:
> LGTM for initial land assuming CI is green

@d4l3k Really appreciate your comments and approval! Could you please help trigger the CI? I don't have the access.

PR #57, opened 2025-11-13
support for testing gloo with cuda tensors

### LINKED ISSUE #54: gloo: support CUDA tensors and verify correctness (CPU reduction, CUDA tensor)
We currently only test Gloo with CPU tensors. ProcessGroup in c10d supports both CUDA and CPU tensors. We have some basic code to support this but it hasn't been robustly tested / verified.

The test configurations in CI are controlled via: https://github.com/meta-pytorch/torchcomms/blob/main/comms/torchcomms/scripts/run_tests_integration_py.sh#L26-L32

Testing:

We want to test with CUDA backend and gloo.

For example to run the alltoall single test:

```sh
TEST_BACKEND=gloo TEST_DEVICE=cuda torchrun --nnodes 1 --nproc_per_node 4 comms/torchcomms/tests/integration/py/AllToAllSingleTest.py
```

We can also test E2E with torchtitan which should give a good indication if things are being handled incorrectly

Things to watch out for:

* making sure that we're copying back to the original device tensor for all operations
* making sure that we correctly synchronize the streams and doing the transfer on the right streams

Not necessarily part of this issue but also relevant:
* enabling ibverbs backend for Gloo
* enabling CUDA/GPU reductions for Gloo - this task is only covering CPU reduction for tensors originally located on the GPU

arkadip-maitra:
Fixes #54

### DIFF
--- a/comms/torchcomms/gloo/TorchCommGloo.cpp
+++ b/comms/torchcomms/gloo/TorchCommGloo.cpp
@@ -1024,8 +1024,8 @@ c10::intrusive_ptr<TorchWork> TorchCommGloo::all_to_all_v_single(
         outputElements.reserve(output_split_sizes.size());
 
         // Calculate number of elements in each dim 0 chunk.
-        auto dim0Numel =
-            inputCPU.numel() / std::max(inputCPU.size(0), static_cast<int64_t>(1));
+        auto dim0Numel = inputCPU.numel() /
+            std::max(inputCPU.size(0), static_cast<int64_t>(1));
         for (auto size : input_split_sizes) {
           inputElements.push_back(static_cast<int64_t>(size) * dim0Numel);
         }


comms/torchcomms/gloo/TorchCommGloo.cpp
```diff
@@ -542,11 +557,7 @@ c10::intrusive_ptr<TorchWork> TorchCommGloo::broadcast(
 
         gloo::broadcast(opts);
 
-        if (tensorCPU.device() != tensor.device()) {
```
d4l3k, line 545: Why remove these checks? This is a noop if it's the same tensor?

comms/torchcomms/gloo/TorchCommGloo.cpp
```diff
@@ -443,12 +457,15 @@ c10::intrusive_ptr<TorchWork> TorchCommGloo::recv(
 
   tracing_->recordEventWithInputOutput("recv", src, {tensor}, {tensor});
 
+  auto originalDevice = tensor.device();
```
d4l3k: we don't seem to use originalDevice anywhere?

d4l3k (commented):
Thanks for sending this! I'll kick off the tests

comms/torchcomms/gloo/TorchCommGloo.cpp
arkadip-maitra: yes it ununsed. removed it.

comms/torchcomms/gloo/TorchCommGloo.cpp
arkadip-maitra: true. i was using .clone() which was not needed. so this check can stay.

d4l3k (approved):
LGTM

arkadip-maitra:
@d4l3k i cant see why lint is failing and there is no lintrunner.toml how do i run lint locally?

comms/torchcomms/gloo/TorchCommGloo.cpp
```diff
@@ -1014,7 +1025,7 @@ c10::intrusive_ptr<TorchWork> TorchCommGloo::all_to_all_v_single(
 
         // Calculate number of elements in each dim 0 chunk.
         auto dim0Numel =
-            input.numel() / std::max(input.size(0), static_cast<int64_t>(1));
+            inputCPU.numel() / std::max(inputCPU.size(0), static_cast<int64_t>(1));
```
d4l3k: Sorry, lint isn't public yet

```diff
-        auto dim0Numel =
-            inputCPU.numel() / std::max(inputCPU.size(0), static_cast<int64_t>(1));
+        auto dim0Numel = inputCPU.numel() /
+            std::max(inputCPU.size(0), static_cast<int64_t>(1));
```

arkadip-maitra: fixed. thx!!

PR #76, opened 2025-12-05
transport: fix transport cmake

d4l3k:
This fixes the transport cmake build linking issues and adds a very basic import test to CI.

Test plan:

```
uv pip install -v -e '.[dev]' --no-build-isolation

python -c "import torchcomms; import torchcomms._transport"
```


amirafzali:
let's smoke test the dunder lib imports? https://github.com/meta-pytorch/torchcomms/blob/main/scripts/smoke_test.py#L8

d4l3k:
@amirafzali holding off on that for now since we may not want to build transport in all builds

PR #238, opened 2026-01-20
[torchcomms] add a few window metadata methods

dolpm:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #239
* #120
* #117
* #116
* #114
* #113
* #112
* #111
* #109
* #118
* #115
* __->__ #238
* #259
* #326

Summary:
metadata methods for dtype,shape,device for usage in meta kernels later.

Differential Revision: [D91187850](https://our.internmc.facebook.com/intern/diff/D91187850)

PR #112, opened 2025-12-30
[torchcomms] add autograd

dolpm:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #435
* #239
* #120
* #116
* #118
* #115
* #114
* #113
* __->__ #112
* #111

Differential Revision: [D91275529](https://our.internmc.facebook.com/intern/diff/D91275529)

PR #111, opened 2025-12-30
[torchcomms] torch.compile post-split

dolpm:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #435
* #239
* #120
* #116
* #118
* #115
* #114
* #113
* #112
* __->__ #111

Differential Revision: [D91275530](https://our.internmc.facebook.com/intern/diff/D91275530)

PR #109, opened 2025-12-30
[torchcomms] param parsing utils

dolpm:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #435
* #239
* #120
* #116
* #118
* #115
* #114
* #113
* #112
* #111
* #419
* __->__ #109

Differential Revision: [D91275531](https://our.internmc.facebook.com/intern/diff/D91275531)

PR #259, opened 2026-01-20
support copy/deepcopy for comm,window,reduceop

dolpm:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #239
* #120
* #117
* #116
* #114
* #113
* #112
* #111
* #109
* #118
* #115
* #238
* __->__ #259
* #326

Summary:
there are cases where i'm copying a graph module and need these to be defined at the very least.

i think this should be fine (and correct) for now, though. deepcopies still do not clone backends, as this is quite costly... but the window WILL clone its associated buffer tensor.

Differential Revision: [D91187820](https://our.internmc.facebook.com/intern/diff/D91187820)

### DIFF
--- a/comms/torchcomms/tests/unit/py/test_copy.py
+++ b/comms/torchcomms/tests/unit/py/test_copy.py
@@ -97,7 +97,7 @@ class TestWindowCopy(unittest.TestCase):
             self.window = self.torchcomm.new_window()
         except RuntimeError:
             self.skipTest(
-                f"Unable to create window using backend {os.getenv("TEST_BACKEND")}."
+                f"Unable to create window using backend {os.getenv('TEST_BACKEND')}."
             )
 
     def tearDown(self):


angelayi (approved):

comms/torchcomms/TorchCommPy.cpp
```diff
@@ -42,6 +42,10 @@ PYBIND11_MODULE(_comms, m) {
           "Create default ReduceOp",
           py::arg("opType"),
           py::call_guard<py::gil_scoped_release>())
+      .def("__copy__", [](const ReduceOp& self) { return self; })
+      .def(
+          "__deepcopy__",
```
d4l3k, line 47: This seems reasonable but we should document why we need to poorly implement deepcopy via some comments.

dolpm: i also have no problem doing an *actual* deepcopy here. with that said, it may introduce more confusion since "new_comm" doesn't always a new backend -- perhaps this is another tihng that we should resolve/clarify?

dolpm:
@dolpm has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D91187820).

dolpm:
@dolpm has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D91187820).

dolpm:
@dolpm has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D91187820).

PR #369, opened 2026-01-23
Parameterize ncclx build folly

jlwass:
Summary: We need to us oss build for folly in docker, parametrizing the folly build so that for feedstock builds it pulls from the env but for docker it builds it

Differential Revision: D91340646


seemethere (approved):

PR #625, opened 2026-02-10
Graceful cleanup when finalize() not called before destruction

d4l3k:
Summary:
Previously, the TorchComm backends (NCCL, NCCLX, RCCL, RCCLX) would log an error and potentially crash when a communicator was destroyed without calling finalize(). This change makes the destructor handle this case gracefully by:

1. Logging a WARNING instead of ERROR to indicate the issue without aborting
2. Properly shutting down the timeout watchdog thread to prevent use-after-free

NOTE: this leaks the nccl communicator since shutting down/abort has been shown to cause errors.

This allows applications to terminate safely if they forget to call finalize(), while still warning about the potential resource leak.

Reviewed By: kapilsh, amirafzali

Differential Revision: D92755212


amirafzali (approved):

PR #674, opened 2026-02-12
ci: remove unnecessary deps torchaudio+torchvision

d4l3k:
We're getting downstream issues from torchvision nightlies not being stable. We don't actually use these dependencies so lets just remove them.

https://github.com/meta-pytorch/torchcomms/actions/runs/21959756498/job/63433728308?pr=659

Test plan:

CI


sramakintel:
@d4l3k Thanks for working on this. I was able to reproduce the issue. The solution is to add `--pre` to all of the pip install commands. 

For example, in `.github/scripts/xpu_test.sh`, solution is `python3 -m pip install --no-deps --pre torch pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir 
`

PR #718, opened 2026-02-18
xpu enabled for perf test

songhappy:
enable perf tests on xpu

### DIFF
--- a/comms/torchcomms/tests/perf/py/perf_test_helpers.py
+++ b/comms/torchcomms/tests/perf/py/perf_test_helpers.py
@@ -89,10 +89,10 @@ def create_tensor(
     return torch.ones(num_elements, dtype=dtype, device=device) * float(rank + 1)
 
 
-def sync_device(device: torch.device) -> None:
-    """Synchronize the device stream if it's a CUDA device."""
-    if device.type == "cuda":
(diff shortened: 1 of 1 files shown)


comms/torchcomms/tests/perf/py/perf_test_helpers.py
```diff
@@ -91,8 +91,8 @@ def create_tensor(
 
 def sync_device(device: torch.device) -> None:
     """Synchronize the device stream if it's a CUDA device."""
```
pkourdis:
```suggestion
    """Synchronize the device stream if it's an accelerator device."""
```

pkourdis (changes requested):

pkourdis (approved):
LGTM

comms/torchcomms/tests/perf/py/perf_test_helpers.py
```diff
@@ -90,9 +90,9 @@ def create_tensor(
 
 
 def sync_device(device: torch.device) -> None:
```
pkourdis: Passing the `device` argument is no longer needed for `sync_device`. You need to remove it and update all the calls to `sync_device`.

pkourdis (changes requested):

PR #845, opened 2026-02-25
Add XCCL C++ unit tests, fix work completion handling, and improve CI

ashokei:
Summary:
- Add mock-based C++ unit tests for the XCCL backend (`XcclMock`, `XpuMock`) so the suite runs without Intel XPU hardware. Coverage includes init/finalize state machine, collectives, point-to-point, reduction-op mapping (SUM / AVG / PreMulSum), failure paths, timeouts, split communicator, hint parsing, bootstrap store exchange, and queue garbage collection — 63 tests across four binaries.
- Fix work completion: `TorchWorkXCCL` now uses base-class `WorkStatus` / `setStatus()` instead of a shadow `state_` field, so pre/post wait hooks observe the correct lifecycle; release tensor references on completion.
- Refactor XCCL error handling: extract `throwAsyncError()`, check `CommState::ERROR` / `TIMEOUT` first in `finalize()`, fix queue GC to track `overall_status` rather than last-seen status so async errors are not overwritten.
- Build: guard `test_utils` (CUDA + folly) behind `if(USE_NCCL OR USE_NCCLX)` so XCCL-only `cmake --build` configurations link cleanly.
- CI: add the C++ test build + `ctest` invocation to `.github/scripts/xpu_test.sh` so mock tests run on every push; share oneAPI / conda / PyTorch setup with the existing Python integration job.


comms/torchcomms/xccl/TorchWorkXCCLQueue.cpp
```diff
@@ -2,8 +2,10 @@
 
 namespace torch::comms {
 
-TorchWorkXCCL::WorkStatus TorchWorkXCCLQueue::garbageCollectLocked() {
-  TorchWorkXCCL::WorkStatus last_status = TorchWorkXCCL::WorkStatus::COMPLETED;
+TorchWork::WorkStatus TorchWorkXCCLQueue::garbageCollectLocked(
+    bool isMainThread) {
+  (void)isMainThread;
```
pkourdis: Looks like `isMainThread` is introduced only to be cast to `void`?

comms/torchcomms/xccl/TorchCommXCCL.cpp
```diff
@@ -675,6 +705,14 @@ c10::intrusive_ptr<TorchWork> TorchCommXCCL::all_reduce(
...
       name_, comm_size_, "all_reduce", rank_, tensor, tensor);
 
+  // oneCCL bug skips premul sum if comm_size is 1, so handle it here
+  // We do this before getOperationStream so that the internal stream
+  // waits for this operation to complete if async_op is true.
+  // TODO: remove this workaround when oneCCL bug is fixed
+  if (comm_size_ == 1 && op == ReduceOp::RedOpType::PREMUL_SUM) {
+    applyPreMulFactor(tensor, op);
+  }
+
```
pkourdis: Why `comm_size==1` for `PREMUL_SUM` needs to be handled differently?

comms/torchcomms/xccl/TorchCommXCCL.cpp
```diff
@@ -707,8 +745,10 @@ c10::intrusive_ptr<TorchWork> TorchCommXCCL::all_reduce(
   // reductions natively.
   const auto maybe_new_op = [&]() -> ReduceOp {
     if (op == ReduceOp::RedOpType::PREMUL_SUM) {
-      c10::StreamGuard guard(stream);
-      applyPreMulFactor(tensor, op);
+      if (comm_size_ != 1) {
+        c10::StreamGuard guard(stream);
+        applyPreMulFactor(tensor, op);
+      }
```
pkourdis: This change needs also justification based on https://github.com/meta-pytorch/torchcomms/pull/845/changes#r2934062773

comms/torchcomms/xccl/TorchWorkXCCLQueue.cpp
ashokei: updated, it was orginally for future use.

comms/torchcomms/xccl/TorchCommXCCL.cpp
ashokei: added additional comments

comms/torchcomms/xccl/TorchCommXCCL.cpp
ashokei: added additional comments

comms/torchcomms/xccl/TorchCommXCCL.cpp
```diff
@@ -277,24 +297,29 @@ void TorchCommXCCL::finalize() {
...
+
+  if (work_status == TorchWork::WorkStatus::NOT_STARTED ||
+      work_status == TorchWork::WorkStatus::INPROGRESS) {
     throw std::runtime_error(
         "WorkQ finalize returned in progress or not started state");
   }
 
   // Update comm_state_ based on the work status
-  if (work_status == TorchWorkXCCL::WorkStatus::TIMEDOUT) {
+  if (work_status == TorchWork::WorkStatus::TIMEDOUT) {
```
zhangxiaoli73, line 452: Should you still keep `TorchWorkXCCL`?

zhangxiaoli73 (approved):

.github/scripts/xpu_test.sh
```diff
@@ -44,5 +44,12 @@ python3 -c "import torch; import torchcomms; print(f'Torch version: {torch.__ver
 python3 -c "import torch;print(\"XPU device available\"); print(torch.xpu.is_available())"
 python3 -c "import torch;[print(f'[{i}]: {torch.xpu.get_device_properties(i)}') for i in range(torch.xpu.device_count())];"
 
+cd torchcomms
+cmake -B build -G Ninja -DBUILD_TESTS=ON -DUSE_XCCL=ON -DUSE_NCCL=OFF -DUSE_NCCLX=OFF -DUSE_TRANSPORT=OFF
```
frost-intel: I'm encountering build errors due to https://github.com/meta-pytorch/torchcomms/pull/947, because it's requiring CUDA.

comms/torchcomms/xccl/TorchCommXCCL.cpp
frost-intel: What is the benefit of moving the comm_size == 1 case to outside this stream guard? Would the same benefit apply to larger comm sizes?

comms/torchcomms/xccl/TorchCommXCCL.cpp
ashokei: The previous TorchWorkXCCL::WorkStatus was a duplicate of TorchWork::WorkStatus with identical values, and state_ was a shadow of the base class's status_. This caused bugs: setStatus() (which triggers post-hook callbacks) was never called, instead we did direct state_ = ... assignments, so TorchComm post-hooks never fired for XCCL work. Removing the duplicate and using the inherited setStatus()/status() fixes callback propagation and aligns with how other backends work.

comms/torchcomms/xccl/TorchCommXCCL.cpp
ashokei: For comm_size == 1, the oneCCL allreduce with PREMUL_SUM is a no-op — it never applies the scaling factor (see https://github.com/uxlfoundation/oneCCL/issues/196). We apply the factor on the current stream (before
  getOperationStream) so that when async_op=true, the dependency event recorded by getOperationStream ensures the internal stream waits for the multiply to complete. This maintains correct ordering without needing an explicit StreamGuard.

For comm_size > 1, we must apply the factor on the operation stream (inside the StreamGuard) because the subsequent allreduce call runs there — the factor and the collective need to be on the same stream to avoid a race. So no, moving it outside wouldn't benefit larger comm sizes; it would actually break correctness.

Without this, PREMUL_SUM gives wrong results for single-rank comms. unit-tests cover this.

.github/scripts/xpu_test.sh
ashokei: pushed a fix to handle build issue

.github/scripts/xpu_test.sh
```diff
@@ -44,5 +44,12 @@ python3 -c "import torch; import torchcomms; print(f'Torch version: {torch.__ver
...
+cmake --build build
```
sramakintel: This file is for python integration tests, do we need to make this addition here? Since we don't have C++ integration tests for the CI

chuanqi129 (approved):
LGTM, based on those new mock-based C++ unit test cases, maybe we can enable xccl backend test on pure CPU runners and gating each PR. How do you think?

sramakintel (approved):

frost-intel (approved):

pkourdis (approved):

d4l3k (approved):
LGTM

PR #881, opened 2026-02-28
Add NanCheckHook to detect NaN/Inf in tensors before collectives (#881)

d4l3k:
Summary:
Catches numerical instability early before it propagates across ranks.

- Expose input_tensor/output_tensor on PreHookArgs in pybind
- Add NanCheckHook class with register/unregister/is_enabled API

Test Plan: - Add unit tests and documentation

Reviewed By: kapilsh

Differential Revision: D94752921

Pulled By: d4l3k


comms/torchcomms/tests/unit/py/test_nan_check.py
```diff
@@ -0,0 +1,177 @@
...
+        hook.unregister()
+        comm.finalize()
+
+    def test_non_float_tensor_passes(self) -> None:
+        """Int tensors skip the check."""
+        comm = self._create_comm("test_int")
+        hook = NanCheckHook()
+        hook.register_with_comm(comm)
+
+        tensor = torch.tensor([1, 2, 3])
```
kapilsh: nit: can we be explicit with dtype?

comms/torchcomms/tests/unit/py/test_nan_check.py
```diff
@@ -0,0 +1,177 @@
...
+
+        self.assertTrue(hook.is_enabled())
+
+        # Clean tensors should pass on both
+        tensor = torch.ones(10)
+        comm1.all_reduce(tensor, torchcomms.ReduceOp.SUM, async_op=False)
+        comm2.all_reduce(tensor, torchcomms.ReduceOp.SUM, async_op=False)
+
+        # NaN should fail on either
+        bad_tensor = torch.tensor([float("nan")])
```
kapilsh, line 149: do we expect any specific order?

comms/torchcomms/tests/unit/py/test_nan_check.py
d4l3k: Done, added explicit `dtype=torch.int32`.

comms/torchcomms/tests/unit/py/test_nan_check.py
d4l3k: No specific order — the test explicitly calls `comm1.all_reduce(bad_tensor, ...)`, so the error deterministically references comm1. Added a comment to clarify.

PR #1108, opened 2026-03-16
Generalize FlightRecorder to for XCCL backend

frost-intel:
This PR generalizes FlightRecorder code for any backend. Labels that were hardcoded to `nccl` now use the backend string.

### DIFF
--- a/comms/torchcomms/hooks/fr/tests/py/FlightRecorderTest.py
+++ b/comms/torchcomms/hooks/fr/tests/py/FlightRecorderTest.py
@@ -1036,14 +1036,14 @@ class TestFlightRecorderHook(unittest.TestCase):
         )
 
         # Verify split operation is recorded
-        split_entries = [e for e in entries if e["profiling_name"] == "nccl:split"]
+        split_entries = [e for e in entries if e["profiling_name"].endswith(":split")]
         self.assertGreater(
             len(split_entries), 0, "Should have recorded the split operation"
         )
 
         # Verify all_reduce operations are recorded (at least 2 - one from each comm)
         all_reduce_entries = [
-            e for e in entries if e["profiling_name"] == "nccl:all_reduce"
+            e for e in entries if e["profiling_name"].endswith(":all_reduce")
         ]
         self.assertGreaterEqual(
             len(all_reduce_entries),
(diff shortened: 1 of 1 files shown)


comms/torchcomms/hooks/fr/FlightRecorder.cpp
```diff
@@ -904,8 +904,9 @@ void FlightRecorderHook::onPreHook(
 
   auto pg_name = std::make_tuple(comm_name, pg_desc);
 
-  // Use "nccl:" prefix as expected by the FR trace analyzer
-  std::string profiling_name = "nccl:" + std::string(opToString(args.name));
+  // Use "<backend>:<op>" format as expected by the FR trace analyzer
+  std::string profiling_name =
+      std::string(pg_desc) + ":" + std::string(opToString(args.name));
```
d4l3k, line 915: Have you verified that the trace analyzer will handle this correctly? We were explicitly setting `nccl:` for non-NCCL backends to ensure it was processed correctly. 

While this is "right" we need to update the ftrace analyzer in torch to support it

comms/torchcomms/hooks/fr/tests/py/FlightRecorderTest.py
```diff
@@ -848,14 +833,13 @@ def test_fr_abort_hook_writes_traces_on_simulated_rank_failure(self) -> None:
         """
         import threading
 
-        backend = os.environ["TEST_BACKEND"]
+        if self.backend == "xccl":
+            self.skipTest("XCCL backend does not support comm abort")
```
d4l3k: can we make this check a reusable decorator similar to `skipIf`? i.e `skipBackend("xccl", "reason")`

d4l3k (commented):
generally this seems pretty reasonable but I would like @tushar00jain to review as well

comms/torchcomms/hooks/fr/tests/py/FlightRecorderTest.py
frost-intel: Done.

comms/torchcomms/hooks/fr/FlightRecorder.cpp
frost-intel: For the most part, I fixed the analyzer previously (see https://github.com/pytorch/pytorch/blob/83e84616031c7a3c0f622d8c2384e81b7eda922b/torch/distributed/flight_recorder/components/types.py#L411)  

The only piece of the ftrace analyzer that's currently nccl-hardcoded is https://github.com/pytorch/pytorch/blob/83e84616031c7a3c0f622d8c2384e81b7eda922b/torch/distributed/flight_recorder/components/utils.py#L605. I'll submit a PR upstream to generalize it, it should be very straightforward since we can just drop the "nccl" and use ":coalesced".

PR #1116, opened 2026-03-16
Remove deprecated header files from RCCLX build

alexk101:
Removed the MultiPeerDeviceTransport.cuh file from the RCCLX CMakeLists, which was deprecated in meta-pytorch/torchcomms#991


comms/rcclx/develop/CMakeLists.txt
```diff
@@ -830,9 +830,7 @@ set(CTRAN_FILES
   comms/ctran/algos/AllToAll/AllToAllvDynamicSplit.cc
   comms/ctran/algos/AllToAll/AllToallvDynamicSplitNonContig.cc
   comms/ctran/algos/AllToAll/AllToAllvImpl.h
-  comms/ctran/algos/AllToAll/fb/compressed/CompressedAllToAllv.cc
   comms/ctran/algos/AllToAll/fb/compressed/CompressedAllToAllv.h
```
nawrinsu: This header is not available and causes a build failure. It should be commented out or removed.

comms/rcclx/develop/CMakeLists.txt
nawrinsu: This header is not available and causes a build failure. It should be commented out or removed.

alexk101: I have already fixed that a few minutes before.

PR #1165, opened 2026-03-19
Align cleanup code on abort for XCCL backend

Kanya-Mo:
Align the implementation introduced in https://github.com/meta-pytorch/torchcomms/pull/625 for XCCL backend. This fixes crash issue when finalize() is not called before destruction.


d4l3k (approved):
LGTM

PR #1309, opened 2026-03-27
xccl: bootstrap race fix + logging and lifetime cleanups

pkourdis:
Three XCCL backend improvements:

- **fix(xccl/bootstrap): wait for rank 0 to publish unique ID before reading it**
  Non-zero ranks were racing with rank 0's write of the unique ID into the store. Added `store_->wait()` with the configured timeout so readers block until the key is present.

- **refactor(xccl): ensure store reference is released in destructor and finalize()**
  Defensively reset `options_.store` in `~TorchCommXCCL()` and `finalize()`. No-op on the happy path, but guarantees the caller's store is released even if `init()` throws before clearing the internal reference.

- **refactor(xccl): add comm context to log lines and move version log to init()**
  Pass `this` to `TC_LOG` throughout so every line is tagged with the communicator (name/rank). Move the XCCL version logging from `DefaultXcclApi::setVersionInfo()` into `TorchCommXCCL::init()` so it prints with comm context, and log (rather than silently discard) a `commDestroy` failure in the destructor.

### DIFF
--- a/comms/torchcomms/xccl/TorchCommXCCL.cpp
+++ b/comms/torchcomms/xccl/TorchCommXCCL.cpp
@@ -228,6 +228,12 @@ TorchCommXCCL::~TorchCommXCCL() {
       xccl_comm_ = nullptr;
     }
   }
+
+  // Release the caller's store in case init() threw before it could drop
+  // our internal reference. No-op on the successful-init path.
+  if (options_.store) {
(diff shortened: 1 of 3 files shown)


pkourdis:
@d4l3k can you please review?

d4l3k (approved):
LGTM

PR #2082, opened 2026-04-15
fix silent data corruption for unsupported BAND/BOR/BXOR reduce ops

arkadip-maitra:
`BAND`, `BOR`, and `BXOR` reduce operations were silently mapped to `ncclSum` across all four backends. Only correctly threw error in xccl. 
Raised same error as PyTorch `ProcessGroupNCCL`.


d4l3k (approved):
LGTM

PR #2104, opened 2026-04-16
Fix alltoallv_dynamic_dispatch crash when called with empty output_tensor_list

arkadip-maitra:
`output_tensor_list[0].numel()` was accessed with no bounds check. every other collective in the file validates list size upfront.

**Repro**
```
ncclx = comm.get_backend_impl()
ncclx.alltoallv_dynamic_dispatch([], output_sizes, input, chunk_sizes, chunk_indices, chunk_count, D, False)
```
**Before**
```
exitcode: -11 (SIGSEGV)
```
**Now**
```
RuntimeError: alltoallv_dynamic_dispatch: output_tensor_list must not be empty
```


d4l3k (approved):
LGTM

PR #2200, opened 2026-04-21
torchcomms/CMakeLists: support building from torch

d4l3k:
This allows overriding paths so we can share the torch copies of gloo/nccl. It also sets nogil to avoid throwing warnings on Python 3.13t/3.14t.

See https://github.com/pytorch/pytorch/pull/181046 for details.


kapilsh (approved):
LGTM!

PR #2205, opened 2026-04-22
Fix transport callsites for new globalRegister/globalDeregister signatures

dsjohns2:
Summary:
D101551580 added a `bool ncclManaged` parameter to `RegCache::globalRegister` and a `bool skipRemRelease` parameter to `RegCache::globalDeregister`, inserting them BEFORE the existing `int deviceId` parameter. Three callsites were passing `cudaDev_` (or `*cudaDevice_`) positionally as what used to be `deviceId`, so after the diff landed the int was silently reinterpreted as a bool.

Fix: pass the new bool parameter explicitly (`false /* ncclManaged */` or `false /* skipRemRelease */`) at each callsite so `cudaDev_` / `*cudaDevice_` again lands in the `deviceId` slot. Preserves the prior behavior for all three files and makes the intent of each argument visible at the call — matching the rest of the new-signature callsites (e.g., `window.cc`).

Reviewed By: tianfengfrank, amirafzali

Differential Revision: D101914693


amirafzali (approved):

PR #2275, opened 2026-04-27
Fix BackendWrapper send/recv passing tag as async_op

arkadip-maitra:
`BackendWrapper::send` and `BackendWrapper::recv` forward the `c10d` tag parameter (an integer) into `TorchCommBackend::send/recv `async_op parameter (a bool)

The fix passes `async_op=true` explicitly and ignores tag, matching PyTorch `ProcessGroupNCCL::send` which names the parameter `int /* unused */` and always operates asynchronously on an internal NCCL stream

The `distwrap` layer (`dw.send/dw.recv`) is unaffected. it calls `TorchComm.send` directly and was already correctly not forwarding the tag parameter.


d4l3k (approved):
LGTM

d4l3k:
JFYI: torchcomms send isn't quite the same as c10d send as we only ever create a single underlying communicator so you may still run into issues with this.

PR #2288, opened 2026-04-27
Make torchcomms backend availability reflect the build config

pkourdis:
Previously `setup.py` registered entry points for every backend unconditionally, so a wheel built with `USE_NCCL=OFF` still had an `nccl` entry point that failed at dlopen time with a confusing `ModuleNotFoundError`. Entry points now mirror the same USE_* flags that gate ext_modules, so the wheel's metadata is an accurate source for which backends were built in.

Add also  two helpers in torchcomms:
- `is_backend_built(backend)`: True iff the wheel was built with the backend's extension.
- `built_backends()`: list of all built-in backend names.

Both read entry-point metadata only, so callers can check backend availability without triggering dlopen or C++ static initializers.

### DIFF
--- a/comms/torchcomms/__init__.py
+++ b/comms/torchcomms/__init__.py
@@ -94,4 +94,5 @@ def is_backend_built(backend: str) -> bool:
 
 
 def built_backends() -> list[str]:
+    """Names of all backends the wheel was built with."""
     return [ep.name for ep in entry_points(group="torchcomms.backends")]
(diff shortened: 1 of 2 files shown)


pkourdis:
Example of usage of the new APIs:
```
>>> import torchcomms
>>> torchcomms.is_backend_built("gloo")
True
>>> torchcomms.is_backend_built("xccl")
True
>>> torchcomms.is_backend_built("ncclx")
False
>>> torchcomms.built_backends()
['dummy', 'gloo', 'xccl']
```

pkourdis:
@d4l3k we can use the `is_backend_built` API to address [your comment](https://github.com/pytorch/pytorch/pull/178533#discussion_r3149727801).

d4l3k (approved):
LGTM

PR #2294, opened 2026-04-28
Fix WorkWrapper c10d status tracking and exception crash

arkadip-maitra:
**Changes**

`BackendWrapper.cpp`: `wait()` and `synchronize()` call `c10d::Work::finish()` to set completed_ and exception_ in the base class. Removed `isCompleted()`, `isSuccess()`, and `exception()` overrides since not needed
`BackendWrapper.hpp`: removed the three override declarations
`TorchCommPy.cpp`: fixed exception() binding on `WorkWrapper` that converts `std::exception_ptr` to a python RuntimeError

**Repro**
```
import warnings, torch
from datetime import timedelta
from torchcomms.device_mesh import init_device_mesh

comm = new_comm("ncclx", torch.device("cuda"), name="dp")
mesh = init_device_mesh(mesh_dim_comms=(comm,), mesh_dim_names=("dp",))
dev = torch.device(f"cuda:{comm.get_rank() % torch.cuda.device_count()}")

pg = mesh.get_group("dp")
work = pg.allreduce([torch.ones(4, device=dev)])
try:
    work.wait(timedelta(seconds=5))
except RuntimeError:
    pass

if comm.get_rank() == 0:
    print(f"is_completed: {work.is_completed()}")
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
        print(f"is_success:   {work.is_success()}")
    print(f"exception:    {work.exception()}")
comm.finalize()
```

**Before**
```
is_completed: False
is_success:   False
exception:    CRASH (TypeError: Unregistered type std::exception_ptr)
```
**After**
```
is_completed: True
is_success:   False
exception:    wait timeout not supported
```

### DIFF
--- a/comms/torchcomms/BackendWrapper.cpp
+++ b/comms/torchcomms/BackendWrapper.cpp
@@ -101,38 +101,36 @@ WorkWrapper::WorkWrapper(
   }
 }
 
-bool WorkWrapper::isCompleted() {
-  return work_->isCompleted();
-}
-bool WorkWrapper::isSuccess() const {
-  // Note: Error state tracking is not implemented. This method returns
-  // isCompleted() as a simplification. The underlying TorchWork does not
-  // expose separate success/error states, so we assume completion implies
-  // success. Callers that need error detection should use try/catch around
-  // wait() instead.
-  return work_->isCompleted();
-}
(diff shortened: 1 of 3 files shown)


d4l3k (approved):
LGTM

PR #2323, opened 2026-04-30
Add static supported function (#2323)

tanquer:
Summary:

Added MultiTransportFactory.supported(transport_type) as a static method. Usage from Python:
```python
  from uniflow import MultiTransportFactory, TransportType

  result = MultiTransportFactory.supported(TransportType.RDMA)
  if result.has_value():
      print("RDMA is supported")
  else:
      print(f"RDMA not supported: {result.error()}")
```

Reviewed By: amirafzali

Differential Revision: D103106965


amirafzali (approved):

PR #2365, opened 2026-05-04
Centralize input validation in TorchComm

pkourdis:
TorchComm is the unified API sitting between the PyTorch bridge (BackendWrapper) and backend implementations (`NCCL`, `RCCL`, `XCCL`, `NCCLX`, `RCCLX`, `Gloo`). It's the single chokepoint every caller goes through and has all the context needed for validation (rank, size, device, tensor args), making it the right place to enforce input contracts once rather than duplicating them across each backend.

Move the following checks out of per-backend implementations into `TorchComm.cpp`:
  - Tensor contiguity, same-numel, and same-dtype
  - tensor/split-list lengths must equal comm_size (`all_gather`, `reduce_scatter`, `all_to_all`, `scatter`, `gather`, and their *_v / *_single variants)
  - Output/input numel ratios against comm_size for *_single variants
  - Divisibility of input numel by comm_size for all_to_all_single
  - Per-rank validation and duplicate-rank detection in subgroup APIs

Removes the `ensureTensorContiguous` helpers from the `NCCL`, `RCCL`, and `XCCL` backends since they are no longer needed.

### DIFF
--- a/comms/torchcomms/TorchComm.cpp
+++ b/comms/torchcomms/TorchComm.cpp
@@ -3,11 +3,62 @@
 #include "comms/torchcomms/TorchComm.hpp"
 #include "comms/torchcomms/TorchCommFactory.hpp"
 
+#include <algorithm>
 #include <atomic>
+#include <set>
 
 namespace torch::comms {
 
 namespace {
+
+void ensureTensorContiguous(const at::Tensor& tensor) {
+  TORCH_CHECK(
+      tensor.is_contiguous(),
+      "Tensor must be contiguous for collective operations");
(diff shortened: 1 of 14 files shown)


pkourdis:
@d4l3k can you please review?

d4l3k:
I'm a little hesitant on this one since we do allow users to access the backends directly when needed

pkourdis:
> I'm a little hesitant on this one since we do allow users to access the backends directly when needed

Thanks for the quick feedback, I was not aware of this usage - closing this PR.

PR #2378, opened 2026-05-05
Fix PRODUCT skip threshold in test_c10d_torchcomms

pkourdis:
The previous threshold (`world_size > 34`) targeted `float32` overflow, but `PRODUCT` fails the exactness check well before that: `world_size!` is only exactly representable in `float32` up to `12!`. From `13!` onward the expected Python int result and the `float32` reduction differ causing `assertEqual` to spuriously fail. Lower the threshold to 12 and update the skip message.


pkourdis:
@d4l3k can you please review?

d4l3k (approved):
Where are you running the tests with world size >12? lgtm

pkourdis:
> Where are you running the tests with world size >12? lgtm

We have an internal cluster where we do validation for large number of ranks.

PR #2380, opened 2026-05-05
Fail XCCL build when oneCCL is missing instead of warning

pkourdis:
Fail XCCL build when oneCCL is missing instead of warning


pkourdis:
@d4l3k can you please review?

d4l3k (approved):

PR #2383, opened 2026-05-05
Reduce python dispatch overhead in distwrap collectives

arkadip-maitra:
Three simple changes to reduce per call python overhead in distwrap layer


d4l3k (approved):
LGTM

PR #2384, opened 2026-05-05
Add ncclCommRevoke, ncclCommShrink, ncclCommGrow, and abort() to rccl/rcclx backends

nikitaxgusev:
## Summary

Add fault-tolerance APIs (revoke, shrink, grow, abort) to the `rccl` and `rcclx` TorchComms backends, mirroring the `nccl`/`ncclx` reconfigure-based recovery path.

**What's included:**

- **`ncclCommRevoke`** – When `enable_reconfigure=true`, the timeout watchdog calls `revokeRcclComm()` / `revokeRcclxComm()` instead of aborting the process. Throws `RuntimeError` instead of `SIGABRT`, allowing recovery via `reconfigure()`.
- **`ncclCommShrink`** – `reconfigure()` diffs old membership against new handles and calls `ncclCommShrink` with `NCCL_SHRINK_ABORT`. Rank renumbering and collectives work on the child communicator.
- **`ncclCommGrow` + `ncclCommGetUniqueId`** – Now wired to the real RCCL 2.29 entry points (gated by `NCCL_VERSION_CODE >= NCCL_VERSION(2, 29, 0)`, matching nccl/ncclx). The reconfigure path gained a `findRankInHandles` helper and a dedicated `new_size == 1` branch that creates a private 1-rank comm without shared-store coordination, without this, `commInitRankConfig` failed with `Invalid rank requested` when every rank shrunk to size 1.
- **`abort()`** – Mirrors `TorchCommNCCL::abort()`: dispatches to revoke when `enable_reconfigure` is set, otherwise aborts.
- **finalize-after-abort fix** – `finalize()` captures `comm_state_` at entry and skips the async error throw if `abort()` already set it to `ERROR`.

## Test results (`TEST_BACKEND=rcclx`, RCCL 2.29.7, after rebase onto current `main`)

| Suite | Ranks | Nodes | Per-rank result | Failures |
|---|---:|---:|---|---:|
| AbortTest        |  32 |  4 | 1 passed, 0 skipped       | 0 |
| AbortTest        |  64 |  8 | 1 passed, 0 skipped       | 0 |
| AbortTest        | 128 | 16 | 1 passed, 0 skipped       | 0 |
| ReconfigureTest  |  32 |  4 | 11 passed, 2 skipped      | 0 |
| ReconfigureTest  |  64 |  8 | 11 passed, 2 skipped      | 0 |
| ReconfigureTest  | 128 | 16 | 11 passed, 2 skipped      | 0 |

**Passed in ReconfigureTest** (rcclx):
- core: `test_reconfigure_basic`, `test_reconfigure_identity`, `test_reconfigure_then_collective`, `test_reconfigure_then_allreduce`, `test_reconfigure_after_abort`
- grow: `test_reconfigure_scale_down_up`, `test_reconfigure_single_to_all`, `test_reconfigure_late`, `test_reconfigure_merge_split`
- shrink: `test_shrink_basic` (now also exercises the post-shrink collectives loop + broadcast previously in `test_shrink_multiple_collectives`), `test_shrink_exclude_middle_rank`

**Skipped** (by design): `test_reconfigure_unsupported_backend`, `test_enable_reconfigure_unsupported_backend` — negative tests for backends that do *not* support `reconfigure()`; rcclx is now supported, so these short-circuit on purpose.

## Repro

```bash
source ~/tc_venv/bin/activate
export LD_PRELOAD=~/rccl-sync-install/lib/librccl.so
export LD_LIBRARY_PATH=~/torchcomms-src/comms/torchcomms:~/rccl-sync-install/lib:$LD_LIBRARY_PATH
export TEST_BACKEND=rcclx MASTER_ADDR=$(hostname -s)

# 32 / 64 / 128 ranks -> adjust -n and the hostfile accordingly (8 GPUs/node).
MPIRUN="mpirun --hostfile <hostfile> -n <N> --npernode 8 \
  --mca btl_tcp_if_include eth0 \
  -x LD_PRELOAD -x MASTER_PORT -x MASTER_ADDR -x TEST_BACKEND -x PATH -x LD_LIBRARY_PATH"

MASTER_PORT=29510 $MPIRUN python3 -m pytest --import-mode=importlib comms/torchcomms/tests/integration/py/AbortTest.py -v
MASTER_PORT=29520 $MPIRUN python3 -m pytest --import-mode=importlib comms/torchcomms/tests/integration/py/ReconfigureTest.py -v
```

### DIFF
--- a/comms/torchcomms/rcclx/RcclxApi.cpp
+++ b/comms/torchcomms/rcclx/RcclxApi.cpp
@@ -1,6 +1,9 @@
 // Copyright (c) Meta Platforms, Inc. and affiliates.
 
 #include "comms/torchcomms/rcclx/RcclxApi.hpp"
+#include <stdexcept>
+
+#include "comms/torchcomms/utils/Logging.hpp"
 
 namespace torch::comms {
 
@@ -47,6 +50,11 @@ ncclResult_t DefaultRcclxApi::commAbort(ncclComm_t comm) {
   return ncclCommAbort(comm);
 }
 
+ncclResult_t DefaultRcclxApi::commRevoke(ncclComm_t comm) {
+  std::lock_guard<std::mutex> lock(api_mutex_);
+  return ncclCommRevoke(comm, 0);
+}
+
 ncclResult_t DefaultRcclxApi::commGetAsyncError(
     ncclComm_t comm,
     ncclResult_t* asyncError) {
@@ -62,6 +70,47 @@ ncclResult_t DefaultRcclxApi::commSplit(
     ncclConfig_t* config) {
   std::lock_guard<std::mutex> lock(api_mutex_);
   return ncclCommSplit(comm, color, key, newcomm, config);
+}
+
+ncclResult_t DefaultRcclxApi::commShrink(
+    ncclComm_t comm,
+    int* excludeRanksList,
+    int excludeRanksCount,
+    ncclComm_t* newcomm,
+    ncclConfig_t* config,
+    int shrinkFlags) {
+  std::lock_guard<std::mutex> lock(api_mutex_);
+  return ncclCommShrink(
+      comm, excludeRanksList, excludeRanksCount, newcomm, config, shrinkFlags);
+}
+
+ncclResult_t DefaultRcclxApi::commGetUniqueId(
+    ncclComm_t comm,
+    ncclUniqueId* uniqueId) {
+  std::lock_guard<std::mutex> lock(api_mutex_);
+  (void)comm;
+  (void)uniqueId;
+  TC_LOG(ERROR)
+      << "ncclCommGetUniqueId API is not implemented in RCCLX backend";
+  return ncclInvalidUsage;
+}
+
+ncclResult_t DefaultRcclxApi::commGrow(
+    ncclComm_t comm,
+    int nRanks,
+    const ncclUniqueId* uniqueId,
+    int rank,
+    ncclComm_t* newcomm,
+    ncclConfig_t* config) {
+  std::lock_guard<std::mutex> lock(api_mutex_);
+  (void)comm;
+  (void)nRanks;
+  (void)uniqueId;
+  (void)rank;
+  (void)newcomm;
+  (void)config;
+  TC_LOG(ERROR) << "ncclCommGrow API is not implemented in RCCLX backend";
+  return ncclInvalidUsage;
 }
 
 ncclResult_t DefaultRcclxApi::commRegister(
--- a/comms/torchcomms/rcclx/RcclxApi.cpp
+++ b/comms/torchcomms/rcclx/RcclxApi.cpp
@@ -88,11 +88,15 @@ ncclResult_t DefaultRcclxApi::commGetUniqueId(
     ncclComm_t comm,
     ncclUniqueId* uniqueId) {
   std::lock_guard<std::mutex> lock(api_mutex_);
+#if NCCL_VERSION_CODE >= NCCL_VERSION(2, 29, 0)
+  return ncclCommGetUniqueId(comm, uniqueId);
+#else
   (void)comm;
   (void)uniqueId;
-  TC_LOG(ERROR)
-      << "ncclCommGetUniqueId API is not implemented in RCCLX backend";
+  TC_LOG(ERROR) << "RCCL version " << NCCL_VERSION_CODE
+                << " does not support ncclCommGetUniqueId API";
   return ncclInvalidUsage;
+#endif
 }
 
 ncclResult_t DefaultRcclxApi::commGrow(
@@ -103,14 +107,19 @@ ncclResult_t DefaultRcclxApi::commGrow(
     ncclComm_t* newcomm,
     ncclConfig_t* config) {
   std::lock_guard<std::mutex> lock(api_mutex_);
+#if NCCL_VERSION_CODE >= NCCL_VERSION(2, 29, 0)
+  return ncclCommGrow(comm, nRanks, uniqueId, rank, newcomm, config);
+#else
   (void)comm;
   (void)nRanks;
   (void)uniqueId;
   (void)rank;
   (void)newcomm;
   (void)config;
-  TC_LOG(ERROR) << "ncclCommGrow API is not implemented in RCCLX backend";
+  TC_LOG(ERROR) << "RCCL version " << NCCL_VERSION_CODE
+                << " does not support ncclCommGrow API";
   return ncclInvalidUsage;
+#endif
 }
 
 ncclResult_t DefaultRcclxApi::commRegister(
--- a/comms/torchcomms/rcclx/TorchCommRCCLXReconfigure.cpp
+++ b/comms/torchcomms/rcclx/TorchCommRCCLXReconfigure.cpp
@@ -103,6 +103,28 @@ struct QuorumInfo {
   size_t newMemberCount = 0;
 };
 
+int findRankInHandles(
+    const std::variant<std::unordered_set<InitHandle>, std::vector<InitHandle>>&
+        handles,
+    const InitHandle& myHandle) {
+  return std::visit(
+      [&](const auto& h) -> int {
+        int result = -1;
+        int idx = 0;
+        for (const auto& handle : h) {
+          if (handle == myHandle) {
+            if (result >= 0) {
+              return -1;
+            }
+            result = idx;
+          }
+          idx++;
+        }
+        return result;
+      },
+      handles);
+}
+
 QuorumInfo findQuorum(
     const std::variant<std::unordered_set<InitHandle>, std::vector<InitHandle>>&
         handles) {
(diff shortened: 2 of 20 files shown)


comms/torchcomms/tests/integration/py/ReconfigureTest.py
```diff
@@ -322,6 +322,166 @@ def test_reconfigure_then_allreduce(self):
...
+            self.backend,
+            self.device,
+            name,
+            enable_reconfigure=True,
+            store=self._get_store_for_comm(),
+            timeout=timedelta(seconds=120),
+        )
+
+        init_handles = self._collect_handles(comm, f"{name}_init")
+        work = comm.reconfigure(
```
yuankaichen-amd, line 256: why do we need to call a reconfigure during initialization?

comms/torchcomms/rcclx/RcclxApi.cpp
```diff
@@ -47,6 +49,11 @@ ncclResult_t DefaultRcclxApi::commAbort(ncclComm_t comm) {
   return ncclCommAbort(comm);
 }
 
+ncclResult_t DefaultRcclxApi::commRevoke(ncclComm_t comm) {
+  std::lock_guard<std::mutex> lock(api_mutex_);
+  return ncclCommRevoke(comm, 0);
```
rishisinhanj, line 55: should we add a guard here as  rccl counterpart in RcclApi.cpp gates them behind an NCCL_VERSION_CODE

comms/torchcomms/tests/integration/py/ReconfigureTest.py
nikitaxgusev: @yuankaichen-amd  The `reconfigure()` call in the helper is intentional and follows the same pattern as `test_reconfigure_then_collective` and `test_reconfigure_then_allreduce`: create a comm, collect all handles, then call `reconfigure()` with those handles before performing any further operations.

The shrink tests need this initial `reconfigure()` call to put the comm into reconfigure mode first. Calling `reconfigure()` again with a reduced handle set, which is the actual shrink step, happens after that.

The helper simply extracts this repeated setup logic to avoid duplication across the three shrink tests.

One more thing: these new tests are additional tests that I added on my own initiative. The NCCL branch doesn’t have them. Do you think it’s okay to integrate them, can we add more for better test coverage?

comms/torchcomms/rcclx/RcclxApi.cpp
nikitaxgusev: @rishisinhanj  The intent here was to align with the NCCLX part. `NcclxApi.cpp` has `commRevoke` and `commShrink` without any `NCCL_VERSION_CODE` guard, which matches what we currently have in `RcclxApi.cpp`.

The version guard pattern exists in `RcclApi.cpp`, which is the plain NCCL/RCCL backend, but `NcclxApi.cpp` doesn’t follow that pattern.

Since the goal here is parity with NCCLX, should we keep it unguarded to match `NcclxApi.cpp`, or would you prefer that we add the guards to stay consistent with `RcclApi.cpp` instead? Thanks!

rishisinhanj: That rationale makes sense, so I think we can leave it then. Everything else LGTM

comms/torchcomms/tests/integration/py/ReconfigureTest.py
rishisinhanj: I think adding more tests for your specific use case makes sense to me for enhanced test coverage, and it is probably safer to keep them rather than delete them. If it is requested to delete them, we can make the change!

nikitaxgusev: Thanks for your feedback, first, I guess it is better to review from Meta's side and then add other tests. Because it is still unclear about the process getting the review from Meta's side.

comms/torchcomms/tests/integration/py/ReconfigureTest.py
```diff
@@ -292,11 +298,170 @@ def test_reconfigure_then_allreduce(self):
...
+        expected = sum(range(1, new_size + 1))
+        self.assertTrue(
+            torch.allclose(tensor, torch.full_like(tensor, expected)),
+            f"AllReduce post-shrink failed: got {tensor[0].item()}, expected {expected}",
+        )
+
+        print(f"[Rank {my_rank}] Shrink {self.world_size} -> {new_size}, allreduce OK")
+        comm.finalize()
+
+    def test_shrink_multiple_collectives(self):
```
d4l3k: can we just add some collectives to run after the existing tests? Not sure this needs to be an entirely separate test

Also doesn't seem like it adds much values over the basic test above

comms/torchcomms/tests/integration/py/ReconfigureTest.py
```diff
@@ -292,11 +298,170 @@ def test_reconfigure_then_allreduce(self):
 
         comm.finalize()
 
+
+    def _create_reconfigured_comm(self, name, uuid):
```
d4l3k, line 244: can we update the other tests to use this as well?

comms/torchcomms/tests/integration/py/ReconfigureTest.py
```diff
@@ -292,11 +298,170 @@ def test_reconfigure_then_allreduce(self):
...
+        """Helper: create a comm and perform initial reconfigure."""
+        import torchcomms
```
d4l3k: pre-existing issue but this import should be top level

comms/torchcomms/rcclx/TorchCommRCCLXReconfigure.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+HandleInfo parseHandle(const InitHandle& handle) {
+  // Format: "rcclx:<rank>:<uuid>:<storeAddr>" or legacy "rcclx:<rank>"
+  auto first = handle.find(':');
+  if (first == std::string::npos) {
+    return {-1, -1, ""};
+  }
+  auto second = handle.find(':', first + 1);
+  if (second == std::string::npos) {
+    // Legacy format: "rcclx:<rank>"
```
d4l3k: can we drop this "legacy format" which has never been used?

d4l3k (approved):
This looks pretty much exactly the same as the nccl impls so generally lgtm, just want to clean up the tests a bit

comms/torchcomms/rcclx/TorchCommRCCLXReconfigure.cpp
nikitaxgusev: @d4l3k  Done, dropped in both `rcclx/TorchCommRCCLXReconfigure.cpp` and `rccl/TorchCommRCCLReconfigure.cpp`

comms/torchcomms/tests/integration/py/ReconfigureTest.py
nikitaxgusev: @d4l3k  Moved `import torchcomms` to the top of the file (line ~20) and removed the twelve inline `import torchcomms`

comms/torchcomms/tests/integration/py/ReconfigureTest.py
nikitaxgusev: @d4l3k  Done, `_create_reconfigured_comm` is now the shared setup for the non-grow reconfigure tests.
The grow-tests (`test_reconfigure_scale_down_up`, `test_reconfigure_single_to_all`, `test_reconfigure_late`, `test_reconfigure_merge_split`) keep their own setup because they exercise a different membership shape (single rank → all ranks, abort-and-rejoin, etc.) where `_create_reconfigured_comm`'s step doesn't apply.

comms/torchcomms/tests/integration/py/ReconfigureTest.py
nikitaxgusev: @d4l3k Agreed, deleted `test_shrink_multiple_collectives` and folded its work into `test_shrink_basic`, which now runs the post-shrink allreduce loop and the broadcast on the same shrunk communicator instead of building a second one.

nikitaxgusev:
@d4l3k Hi! Thanks for the review and for your time.

I addressed your review comments. Could you please take another look?

Also, once I get your approval and the PR is ready for merge, could you please let me know what the next steps are from my side to get it merged?

Thanks!

nikitaxgusev:
Hi @d4l3k and @srinathb-meta,

I am sorry that I am mixing issues here, but I don't have direct contact with you. That's why post it here.

The OSS CMake build for the RCCLX backend currently fails (seems after #2337) when linking against
`rcclx-stable` (or any non-dev variant): `comms/torchcomms/rcclx/CMakeLists.txt`
globs every `*.cpp`, which sweeps in `RcclxApiShardedRelay.cpp` (uses
`ncclx::pFree`, `ncclShardedRelayMultiGroupAllReduce`, etc.) alongside its
intended-mutually-exclusive stub `RcclxApiShardedRelayStub.cpp`. BUCK picks one
or the other via `select()` on `rccl_constraint`, but OSS CMake compiles both
and the dev TU then fails to parse against the stable headers.

The fix is a one-line marker-header check that mirrors what the NCCLX
extension already does for its dev-only sources
(`if(EXISTS "${NCCLX_INCLUDE}/nccl_device/core.h")` in
`comms/torchcomms/ncclx/CMakeLists.txt`): probe the same marker under
`RCCLX_INCLUDE` and `list(FILTER … EXCLUDE)` the TU that won't compile,
patch and verified-both-branches build below. 

Would it possible to review that as a separate PR?

```diff
--- a/comms/torchcomms/rcclx/CMakeLists.txt
+++ b/comms/torchcomms/rcclx/CMakeLists.txt
@@ -13,6 +13,30 @@ message(STATUS "RCCLX_INCLUDE is set to ${RCCLX_INCLUDE}")
 set(RCCLX_SHARED_LIB $ENV{RCCLX_LIB}/librccl.so)
 message(STATUS "RCCLX_SHARED_LIB is set to ${RCCLX_SHARED_LIB}")

+if(EXISTS "${RCCLX_INCLUDE}/nccl_device/core.h")
+    message(STATUS
+        "rcclx-dev API headers found, linking RcclxApiShardedRelay.cpp")
+    list(FILTER TORCHCOMMS_RCCLX_SOURCES
+        EXCLUDE REGEX "RcclxApiShardedRelayStub\\.cpp$")
+else()
+    message(STATUS
+        "rcclx-dev API headers NOT found,"
+        " linking RcclxApiShardedRelayStub.cpp")
+    list(FILTER TORCHCOMMS_RCCLX_SOURCES
+        EXCLUDE REGEX "RcclxApiShardedRelay\\.cpp$")
+endif()
+
 add_library(torchcomms_comms_rcclx MODULE ${TORCHCOMMS_RCCLX_SOURCES})
 set_target_properties(torchcomms_comms_rcclx PROPERTIES
     PREFIX ""
```

Branch:
https://github.com/nikitaxgusev/torchcomms/tree/users/nikitaxgusev/rcclx-cmake-shared-relay-autodetect

Compare against `meta-pytorch/main`:
https://github.com/meta-pytorch/torchcomms/compare/main...nikitaxgusev:torchcomms:users/nikitaxgusev/rcclx-cmake-shared-relay-autodetect

Verified by running `cmake` configure twice:

```
# RCCLX_INCLUDE = rccl-sync install (no nccl_device/core.h)
-- rcclx-dev API headers NOT found, linking RcclxApiShardedRelayStub.cpp
build.ninja TUs: RcclxApi.cpp, RcclxApiShardedRelayStub.cpp

# RCCLX_INCLUDE = include tree with nccl_device/core.h marker
-- rcclx-dev API headers found, linking RcclxApiShardedRelay.cpp
build.ninja TUs: RcclxApi.cpp, RcclxApiShardedRelay.cpp
```

**How to reproduce (no patch, off current `upstream/main` 157cc32f8):**

```bash
git clone https://github.com/meta-pytorch/torchcomms.git && cd torchcomms

# Point at any rcclx-stable / rccl-sync install (i.e. an RCCL build without
# the rcclx-dev ncclx::* extensions in rccl.h). On my box:
export ROCM_HOME=/opt/rocm
export RCCL_INCLUDE=/path/to/rccl-sync-install/include/rccl
export RCCL_LIB=/path/to/rccl-sync-install/lib
export RCCLX_INCLUDE=$RCCL_INCLUDE
export RCCLX_LIB=$RCCL_LIB

USE_NCCL=OFF USE_NCCLX=OFF USE_GLOO=OFF \
USE_RCCL=ON USE_RCCLX=ON \
USE_TRANSPORT=OFF USE_TRITON=OFF \
  pip install --no-build-isolation -v .
```

CMake configures fine; the build then fails when the rcclx extension target
compiles `RcclxApiShardedRelay.cpp` against the stable `rccl.h`:

```
comms/torchcomms/rcclx/RcclxApiShardedRelay.cpp:30:3: error:
    'ncclx' has not been declared
   30 |   ncclx::Hints ncclxHints;
      |   ^~~~~
comms/torchcomms/rcclx/RcclxApiShardedRelay.cpp:34:10: error:
    'ncclx' has not been declared
   34 |   return ncclx::allGatherInit(
      |          ^~~~~
comms/torchcomms/rcclx/RcclxApiShardedRelay.cpp:43:10: error:
    'ncclx' has not been declared
   43 |   return ncclx::allGatherExec(sendbuff, count, datatype, request);
      |          ^~~~~
comms/torchcomms/rcclx/RcclxApiShardedRelay.cpp:47:10: error:
    'ncclx' has not been declared
   47 |   return ncclx::pFree(request);
      |          ^~~~~
comms/torchcomms/rcclx/RcclxApiShardedRelay.cpp:61:10: error:
    'ncclShardedRelayMultiGroupAllReduce' was not declared in this scope;
    did you mean 'shardedRelayMultiGroupAllReduce'?
   61 |   return ncclShardedRelayMultiGroupAllReduce(
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |          shardedRelayMultiGroupAllReduce
```

Thanks,
Nikita

srinathb-meta:
@nikitaxgusev Will take a look at the CMake failures shortly.

nikitaxgusev:
@srinathb-meta if you can confirm the fail and you agree on fix, I can create pr, let me know.

srinathb-meta:
@nikitaxgusev There were a couple of other stuff broken in there as well. I landed two PRs to address this (2760 and 2778) a little while ago. I will kick off importing your changes + internal validations shortly.

PR #2389, opened 2026-05-05
Fix garbageCollectLocked status overwrite across streams

arkadip-maitra:
`garbageCollectLocked` iterates multiple stream queues and blindly overwrites `last_status` on every stream, when a completed stream is visited after pending stream, last_status is completed when its not completed. fixed that for nccl, ncclx, xccl. rcc, rcclx use different design.

### DIFF
--- a/comms/torchcomms/nccl/TorchWorkNCCLQueue.cpp
+++ b/comms/torchcomms/nccl/TorchWorkNCCLQueue.cpp
@@ -20,7 +20,9 @@ TorchWorkNCCL::WorkStatus TorchWorkNCCLQueue::garbageCollectLocked() {
 
       // Use the checkStatus function to determine the work status
       TorchWorkNCCL::WorkStatus status = work->checkStatus();
-      last_status = status;
+      if (status != TorchWorkNCCL::WorkStatus::COMPLETED) {
+        last_status = status;
+      }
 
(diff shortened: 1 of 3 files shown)


comms/torchcomms/nccl/TorchWorkNCCLQueue.cpp
```diff
@@ -20,7 +20,9 @@ TorchWorkNCCL::WorkStatus TorchWorkNCCLQueue::garbageCollectLocked() {
 
       // Use the checkStatus function to determine the work status
       TorchWorkNCCL::WorkStatus status = work->checkStatus();
-      last_status = status;
+      if (status != TorchWorkNCCL::WorkStatus::COMPLETED) {
+        last_status = status;
```
d4l3k: Can we move this into the else block below?

d4l3k (changes requested):

comms/torchcomms/nccl/TorchWorkNCCLQueue.cpp
arkadip-maitra: thanks! moved into else block

d4l3k:
@arkadip-maitra do you mind rebasing so CI passes?

d4l3k (approved):
LGTM

PR #2413, opened 2026-05-06
Add Segment.from_tensor() with runtime PyTorch detection

lilyjanjigian:
Summary:
Adds a `Segment.from_tensor(t)` static method to the Python binding that builds a uniflow Segment from a torch.Tensor. Uses runtime detection — no link/import dep on torch — so callers without PyTorch installed are unaffected.

Behavior:
- `import torch` is attempted lazily inside the method. Raises a clear error if torch is unavailable.
- Validates the argument is a contiguous `torch.Tensor`.
- Extracts `data_ptr()`, `numel() * element_size()`, and derives `MemoryType` (cuda → VRAM, cpu → DRAM) and `device_id` from `tensor.device`.

Motivation: torchstore consumes uniflow's Python API directly with PyTorch tensors. `from_tensor()` removes the boilerplate of pulling out the data pointer / length / device manually at every call site.

Reviewed By: saifhhasan, amirafzali

Differential Revision: D104114954


amirafzali (approved):

PR #2419, opened 2026-05-07
return work from alltoallv_dynamic_combine in distwrap

arkadip-maitra:
when `async_op=True` `distwrap.alltoallv_dynamic_combine` was returning None in this wrapper.

```
work = torchcomms.distwrap.alltoallv_dynamic_combine(..., async_op=True)
print(work)   # before: None, after: TorchWork
work.wait()   # before: AttributeError on None; after: works
```


arkadip-maitra:
closed cus distwrap is being removed

PR #2452, opened 2026-05-08
negotiate RDMA QP count (#2452)

tanquer:
Summary:

After MultiTransport intersects CPU and GPU transport factories, RDMA endpoints can still disagree on QP count because the factory config derives numQps from the local NIC selection.

CPU can select eight NICs while a GPU-local factory selects one, and the old topology only advertised the RDMA version, so bind/connect failed later with a QP count mismatch. Add one topology byte for the local QP count, validate that the peer advertises a nonzero count, and create the transport with the common/min QP count so both endpoints bind paired RC QPs without expanding the narrower endpoint. Also reject zero local numQps at construction/topology serialization.

Differential Revision: D104324880

### DIFF
--- a/comms/uniflow/transport/rdma/RdmaTransport.cpp
+++ b/comms/uniflow/transport/rdma/RdmaTransport.cpp
@@ -19,8 +19,19 @@ namespace uniflow {
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


amirafzali (approved):

PR #2468, opened 2026-05-09
Fix torchcomms non root c10d gather handling in BackendWrapper::gather()

arkadip-maitra:
the standard call `dist.gather(tensor, gather_list=gather_list, dst=0)` failed on non root ranks, this bug also effects TorchTitan checkpoint path under `--comm.mode=torchcomms`. 
 
 these are fixed now, also enabled the c10d gather test


kapilsh (approved):
LGTM!

arkadip-maitra:
@kapilsh can you run ci and merge, i dont have that permission

PR #2520, opened 2026-05-13
Update FakeTensorUpdater initialization

benjaminglass1:
Updates for compatibility with https://github.com/pytorch/pytorch/pull/159523.

### DIFF
--- a/comms/torchcomms/functional/passes.py
+++ b/comms/torchcomms/functional/passes.py
@@ -17,6 +17,7 @@ NOTE: when used in conjunction, the reinplacement pass should be applied first.
 
 import logging
 import operator
+from inspect import signature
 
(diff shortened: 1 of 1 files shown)


d4l3k (changes requested):
We run torchcomms against both nightly and stable so this needs to be able to handle both behaviors in order to keep CI passing

benjaminglass1:
@d4l3k I think I got that working, although the CI on this PR will for sure be broken until the PyTorch PR goes in.

benjaminglass1:
@d4l3k Actually, came up with a way to get this working across versions for now. This should actually be mergeable.

d4l3k (approved):
LGTM

PR #2521, opened 2026-05-13
Add pure-Python symmem backend using SymmetricMemory + Triton

d4l3k:
Registers a new "symmem" backend that implements core collectives entirely in Python on top of torch._distributed._symmetric_memory (CUDA P2P/cudaIpc) and a generic Triton all-reduce kernel. No NCCL/NCCLX/Gloo dependency for the collective data path; the gloo PG is only used by CUDASymmetricMemory for rendezvous-time IPC handle exchange.

- Workspace: one shared symmetric buffer per WORLD group (128 MiB default, hint symmem_workspace_bytes).
- Reductions: fast path via torch.ops.symm_mem.one_shot_all_reduce (fp32/bf16+SUM); fallback via a Triton kernel that reads each peer's buffer through buffer_ptrs_dev and reduces in registers.
- broadcast/all_gather/all_to_all/scatter/gather: stage -> barrier ->
  peer-buffer copies via symm_mem.get_buffer.
- send/recv: per-peer tail slots + signal-pad channels (unique per comm).
- Sub-groups share WORLD's workspace and sync via a deterministic signal-pad channel (CUDASymmetricMemory can only allocate on one group per process).

Exposes ReduceOp.factor in the C++ pybind so Python backends can read the PREMUL_SUM factor.

### DIFF
--- a/comms/torchcomms/symmem/triton_kernels.py
+++ b/comms/torchcomms/symmem/triton_kernels.py
@@ -1,0 +1,168 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+
+# pyre-unsafe
+"""Triton kernels backing the symmem backend.
+
+The kernels read directly from peers' symmetric-memory buffers using the
+device-side pointer table returned by ``_SymmetricMemory.buffer_ptrs_dev``
+and apply reductions / element-wise operations in shared/register memory.
+
+The host issues a ``_SymmetricMemory.barrier`` (or a signal-pad-based
+sub-group barrier) around the kernel launch so that the kernel only sees
+consistent peer data.
+"""
+
+from __future__ import annotations
+
+import triton
+import triton.language as tl
+
+
+# Op codes passed as a ``tl.constexpr`` to specialize each kernel.
+# Wrapped in tl.constexpr so they can be referenced from @triton.jit
+# functions (Triton disallows access to plain Python globals).
+_OP_SUM = tl.constexpr(0)
+_OP_PROD = tl.constexpr(1)
+_OP_MIN = tl.constexpr(2)
+_OP_MAX = tl.constexpr(3)
+_OP_AVG = tl.constexpr(4)
+_OP_PREMUL_SUM = tl.constexpr(5)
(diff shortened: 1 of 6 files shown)


githubsgi:
@d4l3k , this will only  work for scale-up, right ?

comms/torchcomms/symmem/triton_kernels.py
```diff
@@ -0,0 +1,168 @@
...
+    # Initialise accumulator with peer 0's contribution; this avoids needing
+    # an explicit identity element (e.g. -inf for MAX, +inf for MIN, 1 for
+    # PROD) which would also need correct typing per dtype.
+    peer0_raw = tl.load(buffer_ptrs_dev + 0)
+    peer0_ptr = tl.cast(peer0_raw, tl.pointer_type(DTYPE))
+    acc = tl.load(peer0_ptr + offs, mask=mask, other=0)
+    if OP == _OP_PREMUL_SUM:
+        acc = (acc.to(tl.float32) * scale).to(DTYPE)
+
+    for r in range(1, world_size):
```
githubsgi, line 92: This is a  sequential loop, not parallelized , right ?

d4l3k: yes, I believe so. This code is a vibe coded proof of concept not performance tuned

cfgfung: > yes, I believe so. This code is a vibe coded proof of concept not performance tuned

We might work together to further enhance the performance.  For instance, kernel fusion is feasible here: 
https://github.com/meta-pytorch/torchcomms/blob/173dede852c9b00901619a76c9c8983392ef1ed9/comms/torchcomms/symmem/backend.py#L622

githubsgi: It also looks  very wasteful  from both compute and communications perspective, as every rank  is copying from every other rank and going through the exact same compute, 

The ring algorithm commonly used by nccl/xccl/rccl etc is  very efficient from both perspective.

PR #2547, opened 2026-05-14
make cuda graph capture match normal failed com behavior

arkadip-maitra:
before, if communicator failed, normal not cuda capture ops would fail fast but cuda captured ops hangs instead of failing, this makes cuda captured ops follow same fail fast as normal path

### DIFF
--- a/comms/torchcomms/nccl/TorchCommNCCL.cpp
+++ b/comms/torchcomms/nccl/TorchCommNCCL.cpp
@@ -92,6 +92,7 @@ void TorchCommNCCL::init(
   device_ = device;
   name_ = name;
   options_ = options;
+  comm_abort_requested_ = false;
 
   if (init_state_ == InitializationState::INITIALIZED) {
     throw std::runtime_error("TorchCommNCCL already initialized");
@@ -267,6 +268,10 @@ void TorchCommNCCL::finalize() {
     throw std::runtime_error("Work timed out during finalize");
   } else if (work_status == TorchWorkNCCL::WorkStatus::ERROR) {
     comm_state_ = CommState::ERROR;
+    if (!nccl_comm_) {
+      throw std::runtime_error(
+          "NCCL communicator was already aborted due to a previous error");
+    }
     ncclResult_t asyncErr;
     NCCL_CHECK(
         nccl_api_,
@@ -333,14 +338,16 @@ void TorchCommNCCL::finalize() {
 }
 
 void TorchCommNCCL::abortNcclComm() {
+  if (comm_abort_requested_.exchange(true)) {
+    return;
+  }
+
   detachMemoryHook();
-  if (nccl_comm_) {
+  ncclComm_t comm = nccl_comm_;
+  nccl_comm_ = nullptr;
+  if (comm) {
     NCCL_CHECK(
-        nccl_api_,
-        nccl_comm_,
-        nccl_api_->commAbort(nccl_comm_),
-        "NCCL Abort failed");
-    nccl_comm_ = nullptr;
+        nccl_api_, comm, nccl_api_->commAbort(comm), "NCCL Abort failed");
   }
   // Never abort the process in reconfigurable mode: callers fall back to
   // revoke + throw so the failure can be handled by reconfiguring.
@@ -349,6 +356,20 @@ void TorchCommNCCL::abortNcclComm() {
     TC_LOG(ERROR, this) << "Aborting process due to timeout";
     runAbortHooks();
     ::abort();
+  }
+}
+
+void TorchCommNCCL::abortNcclCommNoThrow() noexcept {
+  if (comm_abort_requested_.exchange(true)) {
+    return;
+  }
+
+  detachMemoryHook();
+  ncclComm_t comm = nccl_comm_;
+  nccl_comm_ = nullptr;
+  if (comm && nccl_api_) {
+    NCCL_CHECK_IGNORE(
+        nccl_api_, nccl_api_->commAbort(comm), "NCCL Abort failed");
   }
 }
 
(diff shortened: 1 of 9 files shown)


dolpm:
sorry - i haven't had much time to look at this yet (i'll try to get to it in the next day or two). although i'm not quite sure what the goal is here after taking a quick peek - captured ops won't call checkAndAbortIfTimedOutOrError (won't get baked into the graph). for this reason, we have TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING

arkadip-maitra:
> sorry - i haven't had much time to look at this yet (i'll try to get to it in the next day or two). although i'm not quite sure what the goal is here after taking a quick peek - captured ops won't call checkAndAbortIfTimedOutOrError (won't get baked into the graph). for this reason, we have TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING

hey, thanks for the comment. this is not for replay time timeout handling. the graph replay monitoring still belongs to `TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING`. this change is only for the case where the comm is already in timeout or error before capture starts. this change makes it fail fast like the normal path, it was skipping this guard before

dolpm:
> > sorry - i haven't had much time to look at this yet (i'll try to get to it in the next day or two). although i'm not quite sure what the goal is here after taking a quick peek - captured ops won't call checkAndAbortIfTimedOutOrError (won't get baked into the graph). for this reason, we have TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING
> 
> hey, thanks for the comment. this is not for replay time timeout handling. the graph replay monitoring still belongs to `TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING`. this change is only for the case where the comm is already in timeout or error before capture starts. this change makes it fail fast like the normal path, it was skipping this guard before

can you add unit tests that show the failure before, and the closed gap after?

arkadip-maitra:
> > > sorry - i haven't had much time to look at this yet (i'll try to get to it in the next day or two). although i'm not quite sure what the goal is here after taking a quick peek - captured ops won't call checkAndAbortIfTimedOutOrError (won't get baked into the graph). for this reason, we have TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING
> > 
> > 
> > hey, thanks for the comment. this is not for replay time timeout handling. the graph replay monitoring still belongs to `TORCHCOMM_NCCLX_GRAPH_TIMEOUT_MONITORING`. this change is only for the case where the comm is already in timeout or error before capture starts. this change makes it fail fast like the normal path, it was skipping this guard before
> 
> can you add unit tests that show the failure before, and the closed gap after?

added test

dolpm (approved):

arkadip-maitra:
hey @dolpm thanks for the approval, this test only created `capture_stream` on rank 1 so in multi rank run, ranks 2 to N failed before exercising the actual test. While fixing this I noticed the possibility of double `commAbort`, so fixed that too

arkadip-maitra:
the failure seems unrelated, i had similar test failure locally without this pr changes

dolpm:
> the failure seems unrelated, i had similar test failure locally without this pr changes

yeah, xpu failure is known. let me retry the other

arkadip-maitra:
> > the failure seems unrelated, i had similar test failure locally without this pr changes
> 
> yeah, xpu failure is known. let me retry the other

failed again with the same `torch.distributed.DistStoreError: wait timeout after 600000ms` i still think its unrelated cus i found similar hang when i ran this test locally

comms/torchcomms/nccl/TorchCommNCCL.cpp
dolpm, line 270: incase this is called after abortNcclComm which nulls it
```suggestion
    comm_state_ = CommState::ERROR;
    if (!nccl_comm_) {
      throw std::runtime_error(
          "NCCL communicator was already aborted due to a previous error");
    }
```

comms/torchcomms/ncclx/TorchCommNCCLX.cpp
dolpm, line 453: same
```suggestion
    comm_state_ = CommState::ERROR;
    if (!nccl_comm_) {
      throw std::runtime_error(
          "NCCLX communicator was already aborted due to a previous error");
    }
```

comms/torchcomms/nccl/TorchCommNCCL.hpp
```diff
@@ -443,6 +448,10 @@ class TorchCommNCCL : public TorchCommBackend,
       const ncclComm_t comm,
       const ncclDataType_t dataType);
   void timeoutWatchdog() noexcept;
```
dolpm, line 451: feels weird to be we throwing exceptions in noexcept (although we do already throw in there.. hmm). perhaps we could make commAbort not throw, anyway (using the _IGNORE check variants) to prevent termination?

dolpm (approved):
okay sorry again for the delay - should be good on the red signals (edit: nvmd there does seem to be a regression from this that i will look into but should resolve before merging)... 2 additional small requests after looking through this again, tho, then we can merge this finally :)

comms/torchcomms/nccl/TorchCommNCCL.hpp
arkadip-maitra: true, added a watchdog only `abortNcclCommNoThrow()` path and switiching the `noexcept` timeout watchdog path to use this instead

arkadip-maitra:
> okay sorry again for the delay - should be good on the red signals (edit: nvmd there does seem to be a regression from this that i will look into but should resolve before merging)... 2 additional small requests after looking through this again, tho, then we can merge this finally :)

Hey @dolpm thanks for the reviews, updated 😄

PR #2567, opened 2026-05-15
requirements: move expecttest to dev reqs

d4l3k:
```
tristanr@devvm5553 ~/v/torchcomms-0.3.0-release [1]> uv pip install --pre torch torchcomms==0.3.0 --index-url https://download.pytorch.org/whl/test/cu130 --force-reinstall
  × No solution found when resolving dependencies:
  ╰─▶ Because expecttest was not found in the package registry and torchcomms==0.3.0+cu130 depends on expecttest, we can conclude that torchcomms==0.3.0+cu130 cannot be used.
      And because only the following versions of torchcomms are available:
          torchcomms<0.3.0
          torchcomms>=0.3.0+cu130
      and you require torchcomms==0.3.0, we can conclude that your requirements are unsatisfiable.
```

### DIFF
--- a/.github/scripts/xpu_test.sh
+++ b/.github/scripts/xpu_test.sh
@@ -28,10 +28,9 @@ export USE_GLOO=OFF
 export USE_TRANSPORT=OFF
 export USE_SYSTEM_LIBS=1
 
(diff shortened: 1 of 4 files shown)


kapilsh (approved):

kapilsh:
LGTM!

PR #2578, opened 2026-05-16
Centralize glog init in TorchCommBackend ctor

pkourdis:
Move `tryTorchCommLoggingInit()` out of each backend's `init()` and into a single `TorchCommBackend::ensureLoggingInit()` called from the base-class constructor and `TorchCommFactory::get()`. The five backends (`NCCL`, `NCCLX`, `RCCL`, `RCCLX`, `XCCL`) and future backends no longer have to call it, and logging is guaranteed to be initialized before any `TC_LOG` site runs (including the factory's own "Loading backend library" log).

Also default `FLAGS_logtostderr=true` when no `GLOG_*` env var is set, so torchrun-style runs see logs on stderr instead of files in `/tmp`. Users can still override via `GLOG_log_dir`, `GLOG_logtostderr`, or `GLOG_alsologtostderr`.

### DIFF
--- a/comms/torchcomms/TorchCommBackend.cpp
+++ b/comms/torchcomms/TorchCommBackend.cpp
@@ -1,0 +1,46 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+#include "comms/torchcomms/TorchCommBackend.hpp"
+
+#include <glog/logging.h>
+
+#include <cstdlib>
+
+// Google glog's api does not have an external function that allows one to check
+// if glog is initialized or not. It does have an internal function - so we are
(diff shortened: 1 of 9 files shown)


pkourdis:
@d4l3k can you please help review?

pkourdis:
@d4l3k any feedback on this one? It also helps resolve the following warning:
```
 WARNING: Logging before InitGoogleLogging() is written to STDERR
 ```

pkourdis:
@d4l3k resurfacing this one in case it was buried. The glog warning is affecting all backends:
https://github.com/meta-pytorch/torchcomms/actions/runs/27574794097/job/81529355238#step:17:2949
If you are planning to address this differently, I can close this PR.

PR #2581, opened 2026-05-17
Fix high_priority_stream access across backends

arkadip-maitra:
`torchcomms.new_comm("nccl", device, "comm", high_priority_stream=True)` was a no op only only `hints={"high_priority_stream": "true"}` worked. 
now `high_priority_stream=True`  enables high priority streams, changed readme and stub.


d4l3k (approved):
LGTM

d4l3k:
rebased

PR #2724, opened 2026-05-28
workaround to use DLE packages

sramakintel:
The removal of `--no-deps` causes DLE specific packages as dependencies to be installed during torch/torchcomms installation. 
However, due to an ongoing issue with `tcmlib` [here](https://github.com/pytorch/test-infra/issues/8115),pip package installation is currently unstable and there is a fix [here](https://github.com/pytorch/test-infra/pull/8120).  Since this change will be reflected only after the PR is merged and in the subsequent nightly wheel, in order to unblock ourselves, here is a workaround

This issue wasn't observed in the past as the packages were already made available via the `DLE_PATH` that is activated during torchcomms build and test process instead of using pip packages

### DIFF
--- a/.github/scripts/xpu_test.sh
+++ b/.github/scripts/xpu_test.sh
@@ -20,9 +20,16 @@ export USE_GLOO=OFF
 export USE_TRANSPORT=OFF
 export USE_SYSTEM_LIBS=1
 
-python3 -m pip install --pre torch pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir
+python3 -m pip install --no-deps --pre torch pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir
 
-cd torchcomms && pip install '.[dev]' --no-build-isolation && cd ..
+cd torchcomms && \
+    # Parse the base dev array directly out of setup.py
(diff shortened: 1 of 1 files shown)


sramakintel:
@d4l3k can you review this PR?

.github/scripts/xpu_test.sh
```diff
@@ -20,9 +20,16 @@ export USE_GLOO=OFF
...
 export USE_SYSTEM_LIBS=1
 
-python3 -m pip install --pre torch pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir
-
-cd torchcomms && pip install '.[dev]' --no-build-isolation && cd ..
+python3 -m pip install --no-deps --pre torch pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir
+
+cd torchcomms && \
+    # Parse the base dev array directly out of setup.py
+    sed -n '/"dev": \[/,/\]/p' setup.py | grep -v -E '"dev"|\[|\]' | sed -e 's/[", ]//g' > requirements-xccl.txt && \
```
d4l3k, line 27: this feels very fragile, if this is a short term upstream issue should we just wait for that?

d4l3k (changes requested):

sramakintel:
closing this. no longer relevant.

PR #2777, opened 2026-06-03
Fix pyyaml dependency for release

kapilsh:
`extractcvars.py` that uses pyyaml is a codegen step. It's invoked by the build tooling — not by pip's dependency resolver — and that tooling installs yaml itself, right before calling it:

```sh
# comms/ncclx/v2_29/maint/oss_build.sh:327
# Install pyyaml if not already installed (required by extractcvars.py)
conda install pyyaml --yes
```

```python
python3 "$CVARS_DIR/extractcvars.py"     # reads nccl_cvars.yaml → writes nccl_cvars.cc
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
 torch
-pyyaml
 packaging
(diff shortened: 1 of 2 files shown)


d4l3k (approved):
LGTM

PR #2885, opened 2026-06-12
xccl: fix batch_isend_irecv deadlock and enable BackendWrapper tests on XPU

pkourdis:
`batch_isend_irecv` with an `irecv` listed before its `isend` deadlocked on `XPU`. Both ranks hung in oneCCL's `onecclGroupEnd`, blocked in the recv-side Level-Zero IPC handle exchange (`pt2pt_fd_mode_exchange` -> `ccl::utils::recv`).

oneCCL's `group_impl::end()` replays grouped ops in enqueue order via a synchronous loop, and a recv op blocks in a host-side handle exchange until its peer's matching send op is reached. The recv pre-pass that would avoid this is gated on multi-thread instances, so with one process per rank only the in-order blocking loop runs. When both peers enqueue recv before send, each blocks on its first recv before reaching its send -> symmetric deadlock.

Issue all sends before any recvs within the group so the sender-side handles are posted before any blocking recv exchange. Relative order within sends and within recvs is preserved, so peer matching is unchanged.

Also release tensor references in `TorchWorkXCCL::wait()` once the stream wait is recorded; the `XPU` caching allocator preserves stream ordering, so this is safe and avoids holding tensors alive past completion.

Make `BackendWrapperCoalescingTest` and `BackendWrapperShutdownTest` device-agnostic (`torch.accelerator` instead of `torch.cuda`) so they run on` XPU`, and enable the `BackendWrapper`, `C10dBatchIsendIrecv`, `WaitBlocking`, and `FlightRecorderFinalizeOpId` tests in the `XCCL` integration suite.

### DIFF
--- a/comms/torchcomms/scripts/run_tests_integration_xccl_py.sh
+++ b/comms/torchcomms/scripts/run_tests_integration_xccl_py.sh
@@ -20,9 +20,13 @@ run_tests () {
         AllToAllSingleTest.py
         AllToAllTest.py
         AllToAllvSingleTest.py
+        BackendWrapperAllGatherAliasedTest.py
+        BackendWrapperCoalescingTest.py
+        BackendWrapperShutdownTest.py
         BarrierTest.py
         BatchSendRecvTest.py
         BroadcastTest.py
+        C10dBatchIsendIrecvTest.py
         C10dTorchCommTest.py
         DDPCommTest.py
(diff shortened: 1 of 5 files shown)


d4l3k (approved):
LGTM

d4l3k:
@pkourdis lint

PR #2935, opened 2026-06-17
Use local rank for implicit device selection

arkadip-maitra:
fix implicit communicator device selection for indexless accelerators to prefer `LOCAL_RANK` instead of global rank modulo when available

**Repro**
```
import torch

new_comm("ncclx", torch.device("cuda"))
```
**before**:
`With RANK=1, LOCAL_RANK=0, and CUDA_VISIBLE_DEVICES=0,1, torchcomms selected cuda:1`
**after**:
`With RANK=1, LOCAL_RANK=0, and CUDA_VISIBLE_DEVICES=0,1, torchcomms selects cuda:0`

brings it in line with PyTorch on multinode runs


pkourdis:
@arkadip-maitra would it be possible to update this PR to include the `PALS` runtype?
```
enum RunType {
  TorchRun,
  SlurmJob,
  Pals,
  UnknownRun,
};
```
using the following env vars:
```
PALS_APID: Unique application identifier
PALS_LOCAL_RANKID: This process is the N'th rank on this host
```
Ref: https://support.hpe.com/hpesc/public/docDisplay?docId=dp00005984en_us&page=user/pbs/PALS.html&docLocale=en_US

pkourdis:
> @arkadip-maitra would it be possible to update this PR to include the `PALS` runtype?
> 
> ```
> enum RunType {
>   TorchRun,
>   SlurmJob,
>   Pals,
>   UnknownRun,
> };
> ```
> 
> using the following env vars:
> 
> ```
> PALS_APID: Unique application identifier
> PALS_LOCAL_RANKID: This process is the N'th rank on this host
> ```
> 
> Ref: https://support.hpe.com/hpesc/public/docDisplay?docId=dp00005984en_us&page=user/pbs/PALS.html&docLocale=en_US

@arkadip-maitra never mind, we will do this in a follow-up PR just for the completeness of the XCCL backend change.

comms/torchcomms/nccl/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ endif()
 add_library(torchcomms_comms_nccl MODULE
     ${TORCHCOMMS_NCCL_SOURCES}
     ${TORCHCOMMS_CUDA_API_SOURCE}
+    comms/utils/RankUtils.cc
```
d4l3k: why is this included only under the NCCL binary? should this be in libtorchcomms instead and just used from nccl backend?

d4l3k (changes requested):
generally this seems fine, just need to fix the cmake linking

d4l3k (approved):

PR #2954, opened 2026-06-20
Fix RdmaRemoteBuffer unpickling in torchcomms._transport

arkadip-maitra:
`RdmaRemoteBuffer` aborts during unpickling with legacy pickle protocols because the internal torchcomms._transport binding did not reconstruct it as a plain value object.

**repro**
```
rb = torchcomms._transport.RdmaMemory(torch.arange(16, dtype=torch.uint8)).to_remote_buffer()
pickle.loads(pickle.dumps(rb, protocol=0))  
# before: aborts with instance allocation failed
# after: returns RdmaRemoteBuffer
```


d4l3k (approved):
LGTM

PR #3055, opened 2026-07-06
Enable BackendWrapperShutdownTest mixed backend for XCCL

frost-intel:
Extends the mixed-backend idempotent shutdown test to also run on Intel XPU (XCCL), previously restricted to NCCL only.

Changes in BackendWrapperShutdownTest.py:
* Broaden the skip guard to run for both nccl and xccl backends instead of nccl only.
* Replace CUDA-specific calls and naming.
* Derive the device type (cuda vs xpu) and backend pair (cpu:gloo,{device}:{backend}) dynamically from TEST_BACKEND, so the process group and device_id are built for the correct accelerator.

No behavioral change for NCCL; the test now exercises the same idempotent destroy_process_group() path on XPU.


sramakintel:
@d4l3k can you review this PR?

d4l3k (approved):

PR #3132, opened 2026-07-14
[ROCm][CI] Enable ROCm CI for ROCm wheel builds

akashveramd:
This PR enables ROCm CI for building ROCm wheels. Restructured build_wheels_linux.yml such that it can support wheel build for both CUDA & ROCm. Also, made changes to build options.


akashveramd:
@atalman: Could you please approve the workflows for this PR?

.github/workflows/build_wheels_linux.yml
```diff
@@ -41,4 +41,3 @@ jobs:
       test-infra-ref: main
       build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
       trigger-event: ${{ github.event_name }}
-      runner: linux.g5.12xlarge.nvidia.gpu
```
jithunnair-amd, line 44: @kapilsh @d4l3k Do we need >64GB SRAM to build torchcomms on CUDA? We tried to use the default [`linux.g5.4xlarge.nvidia.gpu`](https://aws.amazon.com/ec2/instance-types/g5/) runner label and it [ran into OOM](https://github.com/meta-pytorch/torchcomms/actions/runs/29515458056/job/87928039329).

.github/workflows/build_wheels_linux.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       test-infra-repository: pytorch/test-infra
       test-infra-ref: main
       with-cuda: enable
-      with-rocm: disable
+      with-rocm: enable
       with-cpu: disable
```
jithunnair-amd, line 26: @huydhn @atalman The [generate_binary_build_matrix.py script](https://github.com/pytorch/test-infra/blob/9308e622d0943d9eaef74da7d7abcf488e59d448/tools/scripts/generate_binary_build_matrix.py#L116) has hardcoded runner labels to use for CUDA/CPU/etc. For the torchcomms case, if we want to override the runner used for CUDA/ROCm, we either:
1. post-process the matrix returned from this step, before passing it to the `build` step below, OR
2. invoke `generate_binary_build_matrix.yml` and `build_wheels_linux.yml` separately for CUDA and ROCm, to allow overriding the default runner for each (as was attempted in [this commit](https://github.com/meta-pytorch/torchcomms/pull/3132/changes/f69db5053fe66416c11b00fc6c9aa725f1f873b2), but it seemed extra legwork)

Do you have a suggestion/preference here?

jithunnair-amd: Option 1 would be something not-that-elegant either:
```
  configure-matrix:
    needs: generate-matrix
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.set-runners.outputs.matrix }}
    steps:
      - id: set-runners
        env:
          MATRIX: ${{ needs.generate-matrix.outputs.matrix }}
        run: |
          patched=$(python3 - <<'PY'
          import json, os
          runners = {
              "cuda": "linux.g5.4xlarge.nvidia.gpu",          # or g5.12xlarge if you prefer
              "rocm": "linux.rocm.gpu.ecosystem.gfx950.4",
          }
          matrix = json.loads(os.environ["MATRIX"])
          for entry in matrix:
              runner = runners.get(entry.get("gpu_arch_type"))
              if runner:
                  entry["validation_runner"] = runner
          print(json.dumps(matrix))
          PY
          )
          echo "matrix=${patched}" >> "$GITHUB_OUTPUT"
```

scripts/_build_wheel.sh
```diff
@@ -31,4 +31,15 @@ pip install pyyaml
...
+if [[ "${CU_VERSION:-}" == rocm* ]]; then
+  export USE_SYSTEM_LIBS=1
+  export USE_NCCL=0
+  export USE_NCCLX=0
+  export USE_GLOO=0
+  export USE_RCCL=1
+  export USE_RCCLX=0
+  export USE_TRANSPORT=OFF
+  export ROCM_HOME=/opt/rocm
+fi
```
jithunnair-amd: @akashveramd Torchcomms has some assumptions baked in about CUDA, that will need to be sorted out first: https://github.com/meta-pytorch/torchcomms/blob/a69e8bf5cd7064889e2cd56524a27f234d7d6ebb/CMakeLists.txt#L47-L49

That `NEEDS_CUDA` (introduced [here](https://github.com/meta-pytorch/torchcomms/commit/e6ae780bbbfa789a9c592313552ceeba76970bf6)) is actually very CUDA-specific, so it shouldn't be set if `USE_RCCL/USE_RCCLX` is set. We probably need a `NEEDS_ROCM` to avoid all the CUDA-specific code.

@kapilsh @d4l3k Would you rather we include those changes in this PR, or a separate PR?

.github/workflows/build_wheels_linux.yml
d4l3k: generally yes, at least for ncclx/ctran it's required, we can likely adjust the number of parallel jobs to help fix this though but needs tuning

scripts/_build_wheel.sh
d4l3k: feel free to combine them

.github/workflows/build_wheels_linux.yml
d4l3k: cc @atalman , this seems fine?

.github/workflows/build_wheels_linux.yml
jithunnair-amd: It's up to you folks, if you'd like to instead use a beefy CPU-only runner instead, since IIUC, GPUs are not exactly required to _build_ the torchcomms wheel (please correct me if I'm wrong)? We [built with the RCCL backend](https://github.com/meta-pytorch/torchcomms/actions/runs/29785224818/job/88497906690?pr=3132) on a linux.2xlarge ([Runner Type: linux.2xlarge, Instance Type: c5.2xlarge](https://github.com/meta-pytorch/torchcomms/actions/runs/29785224818/job/88497906690?pr=3132#step:2:5)) ([16GB RAM](https://aws.amazon.com/ec2/instance-types/c5/)). Should we enable RCCLX and GLOO backends as well for the ROCm build job?

scripts/_build_wheel.sh
kapilsh: yeah - feel free to combine them unless you think the change is isolated enough. up to you.

.github/workflows/build_wheels_linux.yml
atalman: Hi @jithunnair-amd and @d4l3k yes option 1 seems like more straight forward here. This should be fine for this use case.  I believe this is first time we need this override, if we end up needing this more then once we can extend ``generate-binary-matrix`` to allow passing runner types.

.github/workflows/build_wheels_linux.yml
```diff
@@ -22,10 +22,36 @@ jobs:
...
+      matrix: ${{ steps.set-runners.outputs.matrix }}
+    steps:
+      - id: set-runners
+        env:
+          MATRIX: ${{ needs.generate-matrix.outputs.matrix }}
+        run: |
+          patched=$(python3 - <<'PY'
+          import json, os
+          runners = {
+              "cuda": "linux.g5.12xlarge.nvidia.gpu",
```
jithunnair-amd, line 41: Let's also specify the runner label being used for ROCm here.

.github/workflows/build_test_rccl.yaml
```diff
@@ -16,11 +23,11 @@ jobs:
       matrix:
         include:
           - name: RCCL
-            runs-on: amd-mi350-runner
+            runs-on: linux.rocm.gpu.ecosystem.gfx950.8
```
jithunnair-amd: Since these runner scalesets are currently *not* registered for `meta-pytorch` org, these jobs will not find any runners. Let's separate this file's changes into another PR, while we try to bring up runner scalesets for the `meta-pytorch` org.

scripts/_build_wheel.sh
jithunnair-amd: Would you be interested in having torchcomms be setup similar to PyTorch, where the CUDA/ROCm part is detected via CMake ([cuda](https://github.com/pytorch/pytorch/blob/0fd154363f384ab6efaeb9c795b5fd0389de317e/cmake/Dependencies.cmake#L44)/[ROCm](https://github.com/pytorch/pytorch/blob/0fd154363f384ab6efaeb9c795b5fd0389de317e/cmake/Dependencies.cmake#L1010)), and then turned off or kept on? And the dependent variables such as USE_RCCL/USE_NCCL would be turned on/off accordingly?

Or would you rather always use the env var method to build for ROCm?

d4l3k: I think it's reasonable to autodetect this, one pain though with pytorch cmake setup is that certain things are decided only in cmake and the sync between the two gets weird. If we can put the detection logic in setup.py it sounds reasonable to me

setup.py
```diff
@@ -193,6 +200,9 @@ def build_cmake(self, ext):
             f"-DUSE_TRANSPORT={flag_str(USE_TRANSPORT)}",
             f"-DUSE_TRANSPORT_CCA_HOOK={flag_str(USE_TRANSPORT_CCA_HOOK)}",
             f"-DUSE_TRITON={flag_str(USE_TRITON)}",
+            f"-DUSE_SYSTEM_LIBS={flag_str(USE_SYSTEM_LIBS)}",
+            f"-DROCM_HOME={ROCM_HOME}",
+            f"-DRCCL_INCLUDE={RCCL_INCLUDE}",
```
jithunnair-amd: These shouldn't be passed in to CMake from here because they can be inferred using LoadHIP.cmake, which is part of the torch installation, and included via [`find_package(Torch)`](https://github.com/meta-pytorch/torchcomms/blob/7b285acae058718c67cdd39308f5f86b261ff2cc/CMakeLists.txt#L136)

setup.py
```diff
@@ -92,6 +93,12 @@ def get_torch_pybind11_include_root(build_temp: pathlib.Path) -> pathlib.Path:
     "USE_TRANSPORT_CCA_HOOK", USE_NCCLX and not IS_ROCM
 )
 USE_TRITON = flag_enabled("USE_TRITON", False)
+# ROCm paths: default to standard /opt/rocm install; override via env if needed.
+USE_SYSTEM_LIBS = flag_enabled("USE_SYSTEM_LIBS", IS_ROCM)
+ROCM_HOME = os.environ.get("ROCM_HOME", "/opt/rocm") if IS_ROCM else ""
+RCCL_INCLUDE = os.environ.get(
+    "RCCL_INCLUDE", f"{ROCM_HOME}/include/rccl" if IS_ROCM else ""
+)
```
jithunnair-amd: Not needed (see comment below)

CMakeLists.txt
```diff
@@ -87,7 +95,10 @@ endif()
 # Set ROOT early — needed by build_ncclx.cmake and backend CMakeLists.
 set(ROOT ${CMAKE_CURRENT_SOURCE_DIR})
 
-set(USE_SYSTEM_LIBS $ENV{USE_SYSTEM_LIBS})
+# USE_SYSTEM_LIBS is passed via -D from setup.py; fall back to env for direct cmake invocations.
+if(NOT DEFINED USE_SYSTEM_LIBS)
+    set(USE_SYSTEM_LIBS $ENV{USE_SYSTEM_LIBS})
+endif()
```
jithunnair-amd: This snippet may not be needed as per comments on setup.py changes.

comms/torchcomms/rccl/CMakeLists.txt
```diff
@@ -3,12 +3,21 @@
 file(GLOB TORCHCOMMS_RCCL_SOURCES "comms/torchcomms/rccl/*.cpp")
 
 set(CONDA_INCLUDE "${CONDA_PREFIX}/include")
-set(ROCM_HOME $ENV{ROCM_HOME})
+# ROCM_HOME and RCCL_INCLUDE are passed via -D from setup.py;
+# fall back to env then defaults for direct cmake invocations.
 if(NOT ROCM_HOME)
-    set(ROCM_HOME "/opt/rocm")
```
jithunnair-amd, line 8: Should be replaced with `set(ROCM_HOME ${ROCM_PATH}` instead, since LoadHip.cmake sets `ROCM_PATH`: https://github.com/pytorch/pytorch/blob/42cfe4b71af8e868379176aeda40e8d1c5e3f854/cmake/public/LoadHIP.cmake#L9

comms/torchcomms/rccl/CMakeLists.txt
```diff
@@ -3,12 +3,21 @@
...
+    set(ROCM_HOME $ENV{ROCM_HOME})
+    if(NOT ROCM_HOME)
+        set(ROCM_HOME "/opt/rocm")
+    endif()
```
jithunnair-amd: This code should be removed, since we do not want to hardcode ROCm paths here, and instead use the logic in LoadHip.cmake

comms/torchcomms/rccl/CMakeLists.txt
```diff
@@ -3,12 +3,21 @@
...
 endif()
 set(ROCM_INCLUDE "${ROCM_HOME}/include")
-set(RCCL_INCLUDE $ENV{RCCL_INCLUDE})
```
jithunnair-amd, line 11: Restore for user setting

comms/torchcomms/rccl/CMakeLists.txt
```diff
@@ -3,12 +3,21 @@
...
+if(NOT RCCL_INCLUDE)
+    set(RCCL_INCLUDE $ENV{RCCL_INCLUDE})
+    if(NOT RCCL_INCLUDE)
+        set(RCCL_INCLUDE "${ROCM_HOME}/include/rccl")
+    endif()
+endif()
```
jithunnair-amd, line 19:
```suggestion
if(NOT RCCL_INCLUDE)
    set(RCCL_INCLUDE "${ROCM_HOME}/include/rccl")
endif()
```

comms/torchcomms/rcclx/CMakeLists.txt
```diff
@@ -3,9 +3,13 @@
...
 set(CONDA_INCLUDE "${CONDA_PREFIX}/include")
-set(ROCM_HOME $ENV{ROCM_HOME})
+# ROCM_HOME is passed via -D from setup.py;
+# fall back to env then default for direct cmake invocations.
 if(NOT ROCM_HOME)
-    set(ROCM_HOME "/opt/rocm")
+    set(ROCM_HOME $ENV{ROCM_HOME})
+    if(NOT ROCM_HOME)
+        set(ROCM_HOME "/opt/rocm")
+    endif()
```
jithunnair-amd: Similar comments as for rcc/CMakeLists.txt

jithunnair-amd (changes requested):

comms/torchcomms/rccl/CMakeLists.txt
akashveramd: Since further include paths are dependent on ROCM_HOME (via ROCM_PATH). Hence, this will be ideal-
```
# ROCM_PATH is set by LoadHIP.cmake, included via find_package(Torch).
if(ROCM_PATH)
    set(ROCM_HOME ${ROCM_PATH})
else()
    message(FATAL_ERROR "ROCM_PATH is not set. Ensure PyTorch is installed and LoadHIP.cmake is available.")
endif()
```

.github/workflows/build_wheels_linux.yml
jithunnair-amd: Leaving it to a future PR if the maintainers would like to change the CUDA runner

comms/torchcomms/rccl/CMakeLists.txt
```diff
@@ -3,12 +3,20 @@
...
 set(CONDA_INCLUDE "${CONDA_PREFIX}/include")
-set(ROCM_HOME $ENV{ROCM_HOME})
-if(NOT ROCM_HOME)
-    set(ROCM_HOME "/opt/rocm")
+# ROCM_PATH is set by LoadHIP.cmake, included via find_package(Torch).
+# User can override via ROCM_HOME environment variable.
+if(ROCM_PATH)
+    set(ROCM_HOME ${ROCM_PATH})
+elseif(DEFINED ENV{ROCM_HOME})
+    set(ROCM_HOME $ENV{ROCM_HOME})
```
jithunnair-amd:
```suggestion
if(DEFINED ENV{ROCM_HOME})
    set(ROCM_HOME $ENV{ROCM_HOME})
elseif(ROCM_PATH)
    set(ROCM_HOME ${ROCM_PATH})
```

jithunnair-amd: Flipping the order to maintain user-override priority,

comms/torchcomms/rcclx/CMakeLists.txt
```diff
@@ -3,9 +3,14 @@
...
-set(ROCM_HOME $ENV{ROCM_HOME})
-if(NOT ROCM_HOME)
-    set(ROCM_HOME "/opt/rocm")
+# ROCM_PATH is set by LoadHIP.cmake, included via find_package(Torch).
+# User can override via ROCM_HOME environment variable.
+if(ROCM_PATH)
+    set(ROCM_HOME ${ROCM_PATH})
+elseif(DEFINED ENV{ROCM_HOME})
+    set(ROCM_HOME $ENV{ROCM_HOME})
+else()
```
jithunnair-amd, line 12: Similar comment like https://github.com/meta-pytorch/torchcomms/pull/3132/changes#r3669985109

jithunnair-amd (approved):
Lgtm, once you make the changes to give the user-defined env first priority

jithunnair-amd:
@kapilsh @d4l3k I think this PR is ready to merge. Please review. ROCm CI built successfully. I saw the same CUDA/XPU failures in another PR too: https://github.com/meta-pytorch/torchcomms/pull/3317/checks?check_run_id=90349637357

setup.py
```diff
@@ -77,13 +77,14 @@ def get_torch_pybind11_include_root(build_temp: pathlib.Path) -> pathlib.Path:
...
 print("Configuration:")
-USE_NCCL = flag_enabled("USE_NCCL", True)
-USE_NCCLX = flag_enabled("USE_NCCLX", True)
-USE_GLOO = flag_enabled("USE_GLOO", True)
-USE_RCCL = flag_enabled("USE_RCCL", False)
+IS_ROCM = hasattr(torch.version, "hip") and torch.version.hip is not None
+# Backend defaults flip automatically based on whether the installed torch is ROCm.
+USE_NCCL = flag_enabled("USE_NCCL", not IS_ROCM)
+USE_NCCLX = flag_enabled("USE_NCCLX", not IS_ROCM)
+USE_GLOO = flag_enabled("USE_GLOO", not IS_ROCM)
```
d4l3k: we're disabling Gloo on rocm builds?

d4l3k (changes requested):
Generally looks good but seems like Gloo is incorrectly disabled?

setup.py
akashveramd: Original code had it disabled, so I kept it as it is. But now I enabled it on ROCm.

d4l3k (approved):
LGTM

comms/torchcomms/rcclx/CMakeLists.txt
```diff
@@ -3,9 +3,14 @@
...
-if(NOT ROCM_HOME)
-    set(ROCM_HOME "/opt/rocm")
+# User can override via ROCM_HOME environment variable; otherwise use
+# ROCM_PATH set by LoadHIP.cmake, included via find_package(Torch).
+if(DEFINED ENV{ROCM_HOME})
+    set(ROCM_HOME $ENV{ROCM_HOME})
+elseif(ROCM_PATH)
+    set(ROCM_HOME ${ROCM_PATH})
+else()
+    message(FATAL_ERROR "ROCM_PATH is not set. Ensure PyTorch is installed and find_package(Torch) is called before this point.")
```
jithunnair-amd, line 13:
```suggestion
    message(FATAL_ERROR "ROCM_HOME/ROCM_PATH env var is not set, and ROCm installation path could not be auto-detected. Ensure PyTorch is installed in your environment.")
```

Since this is a user-facing message, it'd be best to only refer to corrective steps they can take on their end, such as defining the env vars, or installing PyTorch (and the `find_package(torch)` should already be done by https://github.com/meta-pytorch/torchcomms/blob/66da3fa70d4963b001af7212b452e02f0ae0b9ca/CMakeLists.txt#L144)

PR #3219, opened 2026-07-21
[XCCL] Remove checks for empty tensors before oneCCL dispatch

newtdms:
When XCCl support was being added for TorchComms, oneCCL did not support zero-sized tensors for several collectives and explicit checks were added to avoid failures. Now that oneCCL supports zero-sized tensors, those checks are redundant and are removed in this patch.


newtdms:
The PR passes internal CI tests.
@pkourdis

sramakintel:
@d4l3k can you review the PR?

d4l3k (approved):
lgtm

d4l3k:
@sramakintel getting xpu test failures

sramakintel:
@d4l3k the errors are not introduced by the PR. The regression in `BackendWrapperShutdownTest` seems to have been caused by this PR in PyTorch- https://github.com/pytorch/pytorch/commit/a4116fb2da that causes regression only with XPU backend. Currently, a portion of this PR fixes the test https://github.com/meta-pytorch/torchcomms/pull/3298. Merging the PR should enable the CI back, verified locally.

sramakintel:
@d4l3k can we merge this PR too?

PR #3499, opened 2026-08-07
Make BackendWrapperShutdownTest work on multi-node again

Kanya-Mo:
To fix the flaky issue, https://github.com/meta-pytorch/torchcomms/pull/2951 introduced use of ephemeral port and publish the port thru node-local file from rank 0. Rank on other hosts cannot find the file so currently the scale-out cases will timeout.  

The PR now uses a single global TCPStore and per-test PrefixStore to avoid rebinding MASTER_PORT in the second test, which also fixes the previous flake issue. At the same time the fix doesn't use any node-local files so the multi-node cases are enabled again.


sramakintel:
@d4l3k can you help review this PR?

d4l3k (approved):
LGTM

PR #3793, opened 2026-08-25
[XCCL] Remove input output tensor size equality condition for all_to_all

newtdms:
Recent changes in torchtitan test cases caused failures for MOE models using TorchComms backend.
When get_spmd_backend() == "spmd_types", it goes through the _AllToAllUneven code path. Current implementation of TorchComms for all_to_all have conditions that does not allow uneven per-rank sizes in the list API. 
This PR address the functional regression and also removes this extra checks to align with NCCL implementation.


newtdms:
The PR is passing integration tests in the internal CI.

PR #4151, opened 2026-09-15
[ROCm][RCCL] Fix bootstrap deadlock after asymmetric subgroup creation

chinmaydk99:
RCCL bootstrap keyed unique-ID exchange with a process-local counter. Members-only communicators advance that counter on participating ranks but not on nonmembers, so a later communicator spanning both sets waits on different store keys and hangs. Derive the key from the communicator name instead, matching the existing NCCL implementation.

This fix is required by [pytorch/pytorch#196907](https://github.com/pytorch/pytorch/pull/196907), which enables c10d/torchcomms RCCL testing on ROCm. That testing exposes this bootstrap deadlock after asymmetric subgroup creation.


pragupta (approved):

d4l3k (approved):
lgtm

